PR #41, opened 2025-03-08
Research user authentication flows with trusted identity providers

JoseSzycho:
This PR adds documentation for Zitadel’s login and registration flows with external identity providers (e.g., GitHub, Google). It covers:

1. **Login Flow for Registered Users**
Details the authentication process for users with an existing account linked to an external IdP, including 2FA if enabled.

2. **Login Flow for Non-Registered Users**
Describes the flow when the user’s account is not registered, including the "User Not Found" warning, registration, email verification, and optional 2FA setup.

3. **Registration Flow**
Explains the registration process for new users via external IdPs, covering authentication, account setup, and optional email verification and 2FA.

4. **User Management Flow**
Explains the process for registering changes to user information or settings on their account.

> [!NOTE]
>
>Each flow is illustrated with sequence diagrams, offering clear steps and optional processes like 2FA and email verification.
>This update provides better clarity for integrating external IdPs with Zitadel.

### DIFF
--- a/enhancements/business-os/identity-and-access-management/authentication/README.md
+++ b/enhancements/business-os/identity-and-access-management/authentication/README.md
@@ -401,7 +401,7 @@ to trust JWT tokens signed by the new authentication system.
 
 ### Login Flow For A Registrated User
 
-This sequence flow illustrates the authentication process using the built-in Zitadel UI when a user logs in via an external identity provider (e.g. GitHub, Google). 
+This sequence flow illustrates the authentication process when a user logs in via an external identity provider (e.g. GitHub, Google). 
 
 > [!NOTE]
 >
@@ -413,7 +413,7 @@ sequenceDiagram
   participant Client as Client
   box rgba(93, 93, 93, 0.07) [IAM System]
   participant IAM as IAM API Server
-  participant IdP as Identity Provider
+  participant IdP as Authentication Provider
   end
   participant EIdP as External Identity Provider
 
@@ -424,8 +424,8 @@ sequenceDiagram
   IdP-->>EIdP: Redirects client to
   User->>EIdP: Authenticates with
   EIdP-->>IdP: Redirects client to
-  opt 2FA is Eneabled
-  User-->>IdP: Perform 2FA
+  opt If additional authentication steps required
+  User-->>IdP: Complete authentication
   end
   IdP-->>IAM: Redirects Client To
   IAM-->>Client: Auth Token
@@ -440,7 +440,7 @@ Diagram explanation:
 
 ### Login Flow For A Non Registrated User
 
-This sequence flow illustrates the authentication process using the built-in Zitadel UI when a user logs in via an external identity provider (e.g. GitHub, Google) and the user account was not previously registered.
+This sequence flow illustrates the authentication process when a user logs in via an external identity provider (e.g. GitHub, Google) and the user account was not previously registered.
 
 > [!NOTE]
 >
@@ -452,7 +452,7 @@ sequenceDiagram
   participant Client as Client
   box rgba(93, 93, 93, 0.07) [IAM System]
   participant IAM as IAM API Server
-  participant IdP as Identity Provider
+  participant IdP as Authentication Provider
   end
   participant EIdP as External Identity Provider
 
@@ -468,8 +468,8 @@ sequenceDiagram
   opt If the registration email differs from the authentication email
   User->>IdP: Verify Email with OTP
   end
-  opt 2FA
-  User->>IdP: Setup 2FA
+  opt Additional authentification
+  User->>IdP: Setup additional authentication
   end
   IdP-->>IAM: Redirects Client To
   IAM-->>Client: Auth Token
@@ -486,7 +486,7 @@ Diagram explanation:
 
 ### Registration Flow
 
-This sequence flow illustrates the registration process using the built-in Zitadel UI when a user logs in via an external identity provider (e.g. GitHub, Google).
+This sequence flow illustrates the registration process when a user logs in via an external identity provider (e.g. GitHub, Google).
 
 ```mermaid
 sequenceDiagram
@@ -494,7 +494,7 @@ sequenceDiagram
   participant Client as Client
   box rgba(93, 93, 93, 0.07) [IAM System]
   participant IAM as IAM API Server
-  participant IdP as Identity Provider
+  participant IdP as Authentication Provider
   end
   participant EIdP as External Identity Provider
 
@@ -509,8 +509,8 @@ sequenceDiagram
   opt If the registration email differs from the authentication email
   User->>IdP: Verify Email with OTP
   end
-  opt 2FA
-  User->>IdP: Setup 2FA
+  opt Additional authentication
+  User->>IdP: Setup additional authentication
   end
   IdP-->>IAM: Redirects Client To
   IAM-->>Client: Auth Token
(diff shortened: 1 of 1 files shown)


enhancements/business-os/identity-and-access-management/authentication/README.md
```diff
@@ -0,0 +1,933 @@
...
+  participant Client as Client
+  participant IdentityProvider as Identity Provider
+  participant IAMSystem as IAM System
+  participant Service as Datum Cloud Service
+
+  User->>Client: Login Request
+  Client->>IAMSystem: Authenticate
+  IAMSystem-->>IdentityProvider: Redirects client to
+  User->>IdentityProvider: Authenticates with
+  IdentityProvider-->>IAMSystem: Redirects client to
```
mustela, line 275: In this diagram, the `IdentityProvider` is an **External** Identity provider right?

enhancements/business-os/identity-and-access-management/authentication/README.md
```diff
@@ -399,6 +399,134 @@ discover the trusted JWKS that are use to validate that JWT authentication
 tokens provided by clients are considered valid. All services should be updated
 to trust JWT tokens signed by the new authentication system.
 
+### Login Flow For A Registrated User
+
+This sequence flow illustrates the authentication process using the built-in Zitadel UI when a user logs in via an external identity provider (e.g. GitHub, Google).
```
mustela, line 404: I would remove “Zitadel” from here since this should really be implementation-agnostic. It doesn’t really matter which auth provider we use; we should be able to maintain the same login flow.

enhancements/business-os/identity-and-access-management/authentication/README.md
```diff
@@ -399,6 +399,134 @@ discover the trusted JWKS that are use to validate that JWT authentication
...
+  participant EIdP as External Identity Provider
+
+  User->>Client: Login Request
+  Client->>IAM: Authenticate
+  IAM-->>IdP: Redirects client to
+  User->>IdP: Select External Identity Provider
+  IdP-->>EIdP: Redirects client to
+  User->>EIdP: Authenticates with
+  EIdP-->>IdP: Redirects client to
+  opt 2FA is Eneabled
```
mustela, line 427: I don’t think we should describe the authentication mechanism the user must use to validate their account since there could be many—OTP, passwordless, multi-factor authentication, etc. These will be handled by the authentication provider.

enhancements/business-os/identity-and-access-management/authentication/README.md
```diff
@@ -399,6 +399,134 @@ discover the trusted JWKS that are use to validate that JWT authentication
...
+  participant EIdP as External Identity Provider
+
+  User->>Client: Login Request
+  Client->>IAM: Authenticate
+  IAM-->>IdP: Redirects client to
+  User->>IdP: Select External Identity Provider
+  IdP-->>EIdP: Redirects client to
+  User->>EIdP: Authenticates with
+  EIdP-->>IdP: Redirects client to
+  Note over IdP: User Not Found Warning
```
mustela, line 466: I don’t think we need or want this. @scotwells, is there any reason why you want to do it this way? I see it’s part of the issue description:

> Users who have not registered should be shown an error that they do not have an account on the platform and should be directed to go through the registration flow.

I don't think there should be any error being displayed to the user. Once the user is authenticated, it will land on our side and we need to "onboard" him if we don't have it yet in our records. So we will probably ask for org/project details, etc. Is that what you are refering to?

scotwells: @mustela this was based on the need to support both a login flow (existing users) and a sign up flow (first time users). @jacobsmith928 mentioned wanting separate flows here to make the UX and analytics clearer.

We also want to make sure that an error would be returned if a user signs up with Google as their identity provider and then attempted to login with GitHub that used the same email. We wouldn't want to automatically create a separate account for the GitHub identity or automatically give the GitHub identity access to an account created for a Google identity.

mustela: @scotwells We can have both sign-in and sign-up options, but if a user tries to sign up with an existing email, I don’t think we should force them to manually switch to the login flow. The system should handle this automatically to ensure a smooth experience. See how Slack handles this when you try to create an account with an existing email.

Screenshot 2025-03-10 at 05 10 44


About using multiples external providers for the same account/email, that's something that [Zitadel can do for us](https://zitadel.com/docs/concepts/features/account-linking). We shouldn't be worried about that.

enhancements/business-os/identity-and-access-management/authentication/README.md
JoseSzycho: In fact, this diagram shows the complete flow from authentication to accessing a protected resource, without a small level of abstraction regarding the authentication steps.

The Identity provider, will have it's implementation that will redirect the user to the external identity provider (Google, GitHub)

enhancements/business-os/identity-and-access-management/authentication/README.md
JoseSzycho: Thanks for the feedback. I also think is going to be better to change the term Internal Identity provider to Authentication Provider to refer to Zitadel and other alternatives. Also going to change that

enhancements/business-os/identity-and-access-management/authentication/README.md
JoseSzycho: I've tried the automatic linking feature. This means, for example, that if you create a username/password account on Zitadel (e.g., user@datum.net), you can later create a new account using GitHub or Google as external identity providers. If the email used with Google matches the existing email (user@datum.net), the user will be asked to link the account to an pre-existing one. This means that only one account is created on Zitadel, and you can log in to that account using different dips.

scotwells: @JoseSzycho does it require that the user authenticate with the username / password used to create their original account before linking a new identity?

JoseSzycho: @scotwells it's not quite clear to me. Let me add a GitHub idp to my local setup, and check if it works with out the need to have a username / password account previously created

JoseSzycho: @scotwells just finished testing. It does not require to authenticate with username / password.  User account can be created, with Google as example, and then you can try to login using GitHub, and then link the accounts.  At Zitadel level, only one account will be created, which has configured two idps for login.

scotwells: @JoseSzycho thanks for confirming that. My concern is that would allow someone to gain unauthorized access to someone's Datum Cloud account if they had originally signed up with Google and then their GitHub account (or some other identity) was compromised. 

Do you know if that behavior can be customized or if there's anything on Zitadel's roadmap to support forcing the user to authenticate with an already linked provider before adding a new one?

JoseSzycho1: @scotwells The account linking is a manual process. If the customer signed up with Google, and now wants sign in with GitHub, they will have to manually link the accounts. For linking an account, an OTP is sent to the email address that the user set up during the account creating with Google. So, if they don't have access to that email account,  the user won't be able to link the account with GitHub. 

**Just finished testing this**

Regarding to:

>Do you know if that behavior can be customized or if there's anything on Zitadel's roadmap to support forcing the user to authenticate with an already linked provider before adding a new one?

Zitadel uses the concept of `PREFERRED LOGIN NAME`, which is the username of the Google or GitHub account that the user used for the account creating.

When a user tries to login, by entering the Login Name manually, if that Login Name matches with a `PREFERRED LOGIN NAME`, the user will be redirected to the corresponding identity provider that was used for signing up the account.

**Just finished testing this**

scotwells: @JoseSzycho was able to walk me through the process of how account linking works in Zitadel and confirmed that a user **must** authenticate with their existing user identity before they can link a new identity. They are also required to have access to the email address used for authentication to perform a OTP exchange. Really happy Zitadel supports this out of the box! 🚀

mustela (approved):
@JoseSzycho LGTM!

@scotwells are you ok with merging this PR and update later with any new finding?

scotwells:
@mustela absolutely! 🚀 ✨

PR #63, opened 2025-03-26
Initial commit of enhancement for edge workload containers (and a pla…

thomasdaly5:
…ceholder for VMs)

### DIFF
--- a/enhancements/compute/workloads/edge-workloads-containers.md
+++ b/enhancements/compute/workloads/edge-workloads-containers.md
@@ -1,0 +1,614 @@
+---
+status: provisional
+stage: alpha|beta|stable
+latest-milestone: "v0.1"
+---
+<!--
+Inspired by https://github.com/kubernetes/enhancements/tree/master/keps/NNNN-kep-template
+
+Goals are aligned in principle with those described at https://github.com/kubernetes/enhancements/blob/master/keps/sig-architecture/0000-kep-process/README.md
+
+Recommended reading:
+  - https://developers.google.com/tech-writing
+-->
+
+<!--
+**Note:** When your Enhancement is complete, all of these comment blocks should be removed.
+
+To get started with this template:
+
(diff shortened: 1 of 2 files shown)


joshlreese:
Would you consider these enhancements to be at a higher level than and complimentary to the existing [Workloads enhancement](https://github.com/datum-cloud/enhancements/tree/main/enhancements/compute/workloads), which gets into a bit more specifics around the API?

thomasdaly5:
> Would you consider these enhancements to be at a higher level than and complimentary to the existing [Workloads enhancement](https://github.com/datum-cloud/enhancements/tree/main/enhancements/compute/workloads), which gets into a bit more specifics around the API?

Tough question: 

On the one hand, a workload is a workload, and hopefully we can use a very similar manifest to define it, so in that regard it would seem like a sub-type to a workload. If that is true and valid, would you want this enhancement stubbed a level down from /compute/workloads/ to something like /compute/workloads/edge?

On the other hand, there is a body of work to be done around connectivity to our anycast network, IPAM, VIPs, and capacity planning, that might justify it sitting at the same level of complexity / effort / scope as our workload operator. 

Am I on the right track in answering your question, or did you have something else in mind?

joshlreese:
> On the other hand, there is a body of work to be done around connectivity to our anycast network, IPAM, VIPs, and capacity planning, that might justify it sitting at the same level of complexity / effort / scope as our workload operator.
> 
> Am I on the right track in answering your question, or did you have something else in mind?

I view the type of content in the enhancements in this PR as more product focused, that would lead to additional enhancements that get into technical detail (in this case, "adopting" the existing workload enhancement, and making more). Maybe it'd make sense to have a `enhancements/compute/workloads/product-design` folder, with each of the enhancements in this PR in their own folders? I'd name the folders the current file names, and have the file name be README.md for some automatic rendering when browsing GitHub.

cc @scotwells - Any thoughts here?

PR #90, opened 2025-05-12
Enhancement for Domains and DNS at Datum

thomasdaly5:
A series of known DNS management problems and related user stories.


enhancements/domains-and-dns/README.md
```diff
@@ -0,0 +1,914 @@
...
+  participant Entri
+  
+  User ->> Portal: Add example.com
+  Portal ->> Entri: Verify example.com
+  Entri ->> User: Modal Dialog for example.com
+  User ->> Entri: Run Entri Domain Ownership Sequence
+  Entri ->> Portal: Domain Ownership Verified
+  Portal ->> User: Domain example.com added to Inventory
+
+```
```
scotwells, line 456: What should this interaction look like if a user creates a Domain via the CLI? Would we tell them to visit the portal to perform the ownership verification or should they be able to verify it without going to the UI? I noticed the **Domain Name Ownership Verification Tool** section above indicates a TXT record could be used for verification.

enhancements/domains-and-dns/README.md
```diff
@@ -0,0 +1,914 @@
...
+spec:
+  name: q7.io
+  registrarIANAID: 1068
+  registrarIANAname: NAMECHEAP INC
+  createdDate: 2012-07-06T13:58:57.00Z
+  modifiedDate: 2019-08-17T16:33:24.36Z
+  expirationDate: 2029-07-06T13:58:57.00Z
+  nameservers: ns1.digitalocean.com, ns2.digitalocean.com, ns3.digitalocean.com
+  DNSSEC: unsigned
+  statusFlags: clientDeleteProhibited, clientTransferProhibited, clientUpdateProhibited
```
scotwells, line 489: We should consider moving some of these into the status of the resource since it's status information that was observed about the domain. The spec of the resource is meant to define the "expected" status of the resource that the customer expects Datum Cloud to work towards. 

```yaml
apiVersion: datumapis.com/v1alpha
kind: Domain
metadata:
  name: q7.io
spec:
  name: q7.io
status:
  registrar:
    id: 1068
    name: NAMECHEAP INC
  creationDate: 2012-07-06T13:58:57.00Z
  modificationDate: 2019-08-17T16:33:24.36Z
  expirationDate: 2029-07-06T13:58:57.00Z
  lastScannedDate: 2025-05-20T12:00:00Z
  nameservers: 
  - ns1.digitalocean.com
  - ns2.digitalocean.com
  - ns3.digitalocean.com
  DNSSEC: unsigned
  statusFlags: 
  - clientDeleteProhibited
  - clientTransferProhibited
  - clientUpdateProhibited
```


We should also use [status conditions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties) to provide status information back to the user on the readiness of their domain.

```yaml
status:
  statusConditions:
  - type: OwnershipVerified
    status: False
    reason: "WaitingForVerification"
    message: "Waiting for user to verify ownership of the domain"
    lastTransitionTime: "2025-05-20T12:00:00Z"
  - type: Ready
    status: False
    reason: "PendingOwnershipVerification"
    message: "Waiting for user to verify ownership of the domain"
    lastTransitionTime: "2025-05-20T12:00:00Z"
```

It would be good if we could brainstorm / enumerate the other status conditions we would want to make available to the user.

enhancements/domains-and-dns/README.md
```diff
@@ -0,0 +1,914 @@
...
+
+Depending upon the capability of Datum's eventing systems, we may want to cause a real time notification to be triggered, but that should be out of scope at this time.
+
+### Domain Inventory View
+
+The domain inventory view will use a table showing all domain names available in the organization.
+
+| Domain Name | Expiration  | Registrar     | DNS Provider   | Status | Last Change Detected | Actions     |
+| ----------- | ----------- | ------------- | -------- | -------------- | ------------------- | ----------- |
+| q7.io       | Jul 6, 2029 | NAMECHEAP INC | Digital Ocean | Normal | Aug 17, 2019         | Details |
```
scotwells, line 527: What does a status of "Normal" indicate here? We should enumerate what health checks need to pass in order for a Domain to have this status and ensure this is surfaced on the status conditions of the Domain.

scotwells (commented):
Overall looks great! Really like the detail on the user stories section.

enhancements/domains-and-dns/README.md
joshlreese: I threw together some thoughts on this topic a couple weeks ago. In the flows shown in the attached image, the Datum API is always the authority of what verification records to expect, which can be validated via multiple mechanisms (manual update & DNS observation, Entri & DNS observation, Entri & Entri webhook).

I didn’t dig too deep into the Entri API at the time, but it did appear that the [entri.showEntri](https://developers.entri.com/api-reference#entri-showentri-config) SDK call lets you provide a set of expected records. If we can get confirmation that this is how it’s expected to be used, then I think we can easily support multiple methods of enabling validation.

IMG_7253

enhancements/domains-and-dns/README.md
thomasdaly5: There is a lot we can do with deciding a domain name's status. To me, a "Normal" status would have the following properties:

- The expiration date is >180 days from the current date.
- The following EPP status flags are set for the domain:  clientDeleteProhibited, clientTransferProhibited, clientUpdateProhibited

There may be edge cases for Registry and Registrars that do not implement all of these EPP code. Per https://www.icann.org/resources/pages/epp-status-codes-2014-06-16-en, we may want to fall back to understand the meaning of "OK".

We might consider having the following matrix of status states:

- Secure: Expiration is >180 dates from current date AND the following EPP status flags are set: serverDeleteProhibited, serverTransferProhibited, serverUpdateProhibited
- Normal: Expiration is >180 dates from current date AND the following EPP status flags are set: clientDeleteProhibited, clientTransferProhibited, clientUpdateProhibited
- Unlocked: Expiration is >180 dates from current date AND none of the above EPP status flags are set. The domain is likely to be "OK".
- Warning: Expiration is < 180 dates from current date OR Any flag other than this un Secure, Normal, or Critical.
- Critical: Anytime any EPP flag of autoRenewPeriod, inactive, pendingDelete, pendingRestore, pendingRenew, pendingTransfer, pendingUpdate, redemptionPeriod, renewPeriod, serverHold, serverRenewProhibited, transferPeriod, clientHold, clientRenewProhibited

enhancements/domains-and-dns/README.md
thomasdaly5: From my reading, Entri does not prescribe the specific records used for domain validation. It is why I suggest the usage of a TXT record with a nonce in the user story. 

In both the API and Portal workflow, we can generate that TXT record, store it as validation metadata, and provide that to the user back over our API to place details of. Could we then use that data to create a status condition for an operator to perform validation checks to the authoritative DNS platform?

If the user is in the Portal workflow, we can also populate the Entri experience to streamline the process for the user. Entri is able to help us place those records with the DNS provider for the domain in question, and we can get a webhook event once those records are placed. We should still perform our own validation.

If this sound acceptable @joshlreese and @scotwells, I will update the flow diagram.

joshlreese: > In both the API and Portal workflow, we can generate that TXT record, store it as validation metadata, and provide that to the user back over our API to place details of. Could we then use that data to create a status condition for an operator to perform validation checks to the authoritative DNS platform?

Yeah, in my sequence diagram the "Update status w/ TXT record requirement" flow implied this. I think it could be a bit more generic, looking something like:

```
apiVersion: networking/datumapis.com/v1alpha1
kind: Domain
metadata:
  name: q7.io
spec: {}
status:
  verification:
    requiredDNSRecords:
      - name: c34ce0b9-a74c-4c6f-8f67-423c682ecbff.q7.io
        type: TXT
        content: <expected value>
```

In the case of the portal creating a domain, it would see these validation requirements and that the `OwnershipVerified` condition is false on the domain, and could prompt the user to go through the Entri flow, using the information in the resource to provide Entri with expected records.

For users creating resources directly, we could link to documentation in the `OwnershipVerified` condition's message for how domain verification works, and the user could either follow manual steps necessary for validation, or log into the portal and go through the Entri route.

Upon successful validation, the system would populate the `OwnershipVerified ` condition similar to what @scotwells suggested in https://github.com/datum-cloud/enhancements/pull/90#discussion_r2098727688.

scotwells: @thomasdaly5 @joshlreese thanks for the additional insight into Entri and how it works. It would be great to add some of that context to the enhancement doc and link off to the Entri docs where folks could read more about how Entri works.

@joshlreese really like that approach for exposing the DNS verification records 👍. That would open the door to exposing it as a printer column on `datumctl`.

enhancements/domains-and-dns/README.md
thomasdaly5: @scotwells I have added your feedback at 3631156827170cfd3ab34bea071b650caefdc74c. Do you think we should be using a set of statusConditions to monitor for these states?

enhancements/domains-and-dns/README.md
thomasdaly5: I've adjusted the proposed manifest per your feedback in 3631156827170cfd3ab34bea071b650caefdc74c.

enhancements/domains-and-dns/README.md
thomasdaly5: Partially updated in 3631156827170cfd3ab34bea071b650caefdc74c. Will adjust flow diagrams further in the next revision.

enhancements/domains-and-dns/README.md
joshlreese: > There may be edge cases for Registry and Registrars that do not implement all of these EPP code. Per https://www.icann.org/resources/pages/epp-status-codes-2014-06-16-en, we may want to fall back to understand the meaning of "OK".

Reading that link, would it be more appropriate to refer to "status flags" as "status codes", and place them under `status.clientStatusCodes` / `status.serverStatusCodes`?

enhancements/domains-and-dns/README.md
```diff
@@ -0,0 +1,978 @@
...
+kind: Domain
+metadata:
+  name: q7.io
+spec:
+  name: q7.io
+    verification:
+      requiredDNSRecords:
+        - name: _datum-cloud-challenge.q7.io
+          type: TXT
+          content: <expected value>
```
joshlreese, line 474: Since our system will be populating this value, it should go into the `status` subresource and not the spec, as we wouldn't want users to need to provide this, or be able to influence it.

enhancements/domains-and-dns/README.md
```diff
@@ -0,0 +1,978 @@
...
+kind: Domain
+metadata:
+  name: q7.io
+spec:
+  name: q7.io
+  verification:
+      requiredDNSRecords:
+        - name: _datum-cloud-challenge_.q7.io
+          type: TXT
+          content: <expected value>
```
joshlreese, line 503: Same comment here regarding where verification info should go - https://github.com/datum-cloud/enhancements/pull/90/files#r2101187938

enhancements/domains-and-dns/README.md
thomasdaly5: As of b6eb5d93272079f459f449383b544f0f36ba9653, I have broken out to two flows:

- A standard flow, which should be command line compatible, in which the user is responsible for providing the domain ownership verification TXT record to their DNS provider out of band.
- A streamlined flow, which calls for use of Entri.

Thank you @joshlreese for contributing your state diagrams.

enhancements/domains-and-dns/README.md
thomasdaly5: Understood. Adjusted. Resolving.

enhancements/domains-and-dns/README.md
thomasdaly5: Understood. Adjusted. Resolving.

enhancements/domains-and-dns/README.md
thomasdaly5: Happy to adjust or leave this for implementation.

joshlreese: Probably best to adjust now so the nomenclature in the enhancement matches that ICANN page, and we won't need to remember this thread of conversation when the time comes to implement this (granted, it'll be soon!).

joshlreese (approved):

PR #99, opened 2025-05-27
Define Galactic VPCs

thomasdaly5:
Add Product Requirements Document (PRD) for NPMS to GitHub. This PRD should be evolved to define Galactic VPC in enhancements format, and can be deleted once completed. No work should be done on NPMS_Full_PRD.md going forward.


mksinghtx:
Adding some reviewers here

jacobsmith928 (changes requested):
@mksinghtx i think we should consider killing this and folding it into the March sprint around Galactic VPC if needed.
@jacobsmith928 I committed my pull rquest - this can be killed now

PR #114, opened 2025-06-05
Service Account Management Diagrams and Flows for Auth Provider Delegation

JoseSzycho:
## Summary

This PR updates the IAM authentication enhancement to clarify and improve the service account lifecycle and key management flows. The changes explicitly model the delegation of service account and key operations to an external Auth Provider (e.g., ZITADEL), and distinguish between user-generated and provider-generated keys.


JoseSzycho:
@scotwells I was thinking of a flow where the private key is returned to the user once at creation time, and never stored on our side. We would store only the public key, keeping the private key fully in the client’s control.

The direction I’m leaning toward is to implement API handlers in Milo for this, and let the `auth provider` handle key generation and public key storage as needed.

PR #205, opened 2025-07-10
Milo Control Plane Architecture

scotwells:
This document covers the initial architectural goals with the Milo Control Plane.


yahyafakhroji (approved):

scotwells:
@ecv no, but I'll update this to include the plantuml equivalent.

PR #233, opened 2025-08-04
feat: e2e testing

zachsmith1:


scotwells:
Can we add canary tests to this that would continuously create resources in the production environment to confirm they're operating as expected? I'd expect something that would go through and provision a HTTP Proxy and confirm it's responding correctly similar to [these end-to-end tests](https://github.com/datum-cloud/datum-infra/tree/main/tests). 

I'd like to move these to an open source repo so we can use it as a way of building trust with the community around how we test our software and the results of those tests.

PR #264, opened 2025-09-02
Milo Performance Testing

zachsmith1:


enhancements/testing/performance/README.md
```diff
@@ -0,0 +1,189 @@
...
+
+### Non-Goals
+
+<!--
+What is out of scope for this Enhancement? Listing non-goals helps to focus discussion
+and make progress.
+-->
+
+- Testing every type of Kubernetes object (we’ll focus on Projects + Secrets/ConfigMaps).  
+- Full multi-tenant end-to-end validation (authn/authz, networking, etc.).
```
scotwells, line 162: @zachsmith1 by this do you mean perf testing won't confirm that authn/authz works as expected? 

I do think there's value in putting authn/authz under load to see how it handles large scale deployments with a large number of organizations, projects, and role bindings, but would agree that's likely separate from testing the multi-tenant aspect of the control plane.

scotwells (approved):

enhancements/testing/performance/README.md
zachsmith1: Ya maybe "validation" wasn't the right word. It might make sense to include some metrics around authn/authz under load here within this scope. We can build upon this later if needed

Issue #224, opened 2025-07-28
Document Agreements and Acceptance Management


JoseSzycho:
### High-Level Summary

We need a way to manage agreements (like Terms of Service) in a versioned, trackable way. This includes storing the agreements, tracking who accepted which version, and making this info visible to both users and staff.

### Motivation

- Ensure we know exactly who agreed to what, and when.
- Make it easy to update agreements over time, while keeping a record of all previous versions.
- Allow for features that require users to accept new terms before using them.
- Build this in a way that’s generic and reusable outside of just our project.

### Goals

- Create a system for defining agreement types and their revisions.
- Store each acceptance event with details (user/org, version, timestamp, IP, etc.).
- Make it possible to view agreement history and acceptance records in both user and staff interfaces.
- Use clear, generic naming so others can adopt this system easily.

### Non-Goals

- No need for full e-signature workflows (like DocuSign).
- No PDF or complex document generation—just Markdown/HTML for now.
- No automated notifications or translations at this stage.

JoseSzycho:
## Usage Examples

### Creating a New Document

```yaml
# Create the document definition
apiVersion: agreements.miloapis.com/v1alpha1
kind: Document
metadata:
  name: privacy-policy
  namespace: legal-documents
spec:
  title: "Privacy Policy"
  description: "How we collect, use, and protect your personal information"
  documentType: "privacy-policy"
  metadata:
    category: "privacy"
    jurisdiction: "US"

---
# Create the first revision
apiVersion: agreements.miloapis.com/v1alpha1
kind: DocumentRevision
metadata:
  name: privacy-policy-v1-0
  namespace: legal-documents
spec:
  documentRef:
    name: privacy-policy
    namespace: legal-documents
  version: "1.0.0"
  content:
    format: "markdown"
    data: |
      # Privacy Policy
      
      Effective Date: January 1, 2025
      
      ## Information We Collect
      ...
    contentHash: "sha256:xyz789abc123def456ghi789jkl012mno345pqr678stu901vwx234yz"
  effectiveDate: "2025-01-01T00:00:00Z"
  changesSummary: "Initial version"
```

### Recording an Acceptance

```yaml
apiVersion: agreements.miloapis.com/v1alpha1
kind: DocumentAcceptance
metadata:
  name: privacy-acceptance-user-67890-20250123
  namespace: organization-techstart
spec:
  documentRevisionRef:
    name: privacy-policy-v1-0
    namespace: legal-documents
[...]

PR #369, opened 2025-10-08
Add Agreements System documentation for Milo platform

JoseSzycho:
Introduces a comprehensive README for the Agreements System, detailing the implementation of a backend management system for legal agreements, including CRDs for Document, DocumentRevision, and DocumentAcceptance. The documentation outlines the system's goals, design details, risks, and future enhancements, emphasizing its focus on versioning and acceptance tracking without a frontend UI.

### DIFF
--- a/enhancements/platform/email-integration/simple-email-sending/README.md
+++ b/enhancements/platform/email-integration/simple-email-sending/README.md
@@ -52,7 +52,7 @@ Milo needs the ability to send emails to end-users based on events happening acr
 
 The Milo platform currently lacks a basic email notification system, which limits our ability to communicate with users about important platform events and notifications. This enhancement addresses the need for a simple, reliable email system for transactional and notification emails.
 
-Modern SaaS platforms require basic email capabilities to notify users about account changes, system events, and provide important service notifications. Without this capability, Milo cannot effectively communicate with users about critical platform events.
+Modern Alt Cloud providers require basic email capabilities to notify users about account changes, system events, and provide important service notifications. Without this capability, Milo cannot effectively communicate with users about critical platform events.
 
 ### Goals
 
@@ -65,6 +65,7 @@ Modern SaaS platforms require basic email capabilities to notify users about acc
 
 **Explicitly out of scope for this enhancement:**
 
+- **Full e-signature workflows** - No built-in support for electronic signature solutions (e.g., DocuSign, Documenso)
 - **Contact management** - No Contact CRDs, contact storage, or contact lifecycle management
 - **Contact groups/mailing lists** - No ContactGroup CRDs or automated group membership
 - **Multi-provider support** - Initial implementation will focus on a single provider (Resend)


enhancements/platform/agreement/README.md
```diff
@@ -0,0 +1,340 @@
...
+  name: terms-of-service
+  namespace: legal-documents
+spec:
+  title: "Terms of Service"
+  description: "The rules and conditions for using our platform."
+  documentType: "terms-of-service"
+  metadata:
+    category: "legal"
+    jurisdiction: "US"
+```
```
scotwells, line 229: What do you think we should show for the status of the document? Maybe a reference to the latest version of the document?

enhancements/platform/agreement/README.md
```diff
@@ -0,0 +1,340 @@
...
+  documentRef:
+    name: terms-of-service
+    namespace: legal-documents
+  version: "2.0.0"
+  content:
+    format: "markdown"
+    data: |
+      # Terms of Service
+      ...
+    contentHash: "sha256:abc123..."
```
scotwells, line 249: Should this be a status field since it's something that's system generated?

enhancements/platform/agreement/README.md
```diff
@@ -0,0 +1,340 @@
...
+  name: tos-acceptance-user-12345-20250602
+  namespace: organization-acme
+spec:
+  documentRevisionRef:
+    name: terms-of-service-v2-0
+    namespace: legal-documents
+    resourceVersion: "12345"
+  userRef:
+    name: user-12345
+  state: "accepted"  # accepted, declined
```
scotwells, line 281: Do we need to track denials?

enhancements/platform/agreement/README.md
```diff
@@ -0,0 +1,340 @@
...
+#### Acceptance Status States
+
+- `accepted` - User has accepted the agreement
+- `rejected` - User declined or failed to accept (optional, for future use)
+
+#### Permission Management
+
+- **On Document/Revision creation:** Only staff/operators can create or update
+- **On Acceptance:** Users can create acceptance events for themselves
+- **On Query:** Staff can view all records; users can view their own
```
scotwells, line 311: We should note how organizations will be able to retrieve their documents.

enhancements/platform/agreement/README.md
```diff
@@ -0,0 +1,340 @@
...
+
+- Full e-signature workflows (e.g., DocuSign)
+- PDF or complex document generation (Markdown/HTML only)
+- Automated notifications or translations
+- Standalone documentation site or SDKs
+- Marketing or product copy
+
+**Future enhancements will address:**
+
+- Frontend UI for agreement management and acceptance
```
scotwells, line 66: Can we expand the scope of this enhancement to include the frontend UI? I think it's fine to do that in a separate PR to keep this one focused on the core API / system.

scotwells (approved):
Off to a great start. Just some minor feedback.

enhancements/platform/agreement/README.md
JoseSzycho: In fact, it is not needed. I was thinking in the case we want to create some metrics. Want me to remove the state field?

enhancements/platform/agreement/README.md
JoseSzycho: Sure, update the enhancement.

Here's the PR I created for tackling that:
https://github.com/datum-cloud/enhancements/issues/224?issue=datum-cloud%7Cenhancements%7C370

enhancements/platform/agreement/README.md
JoseSzycho: That's a great idea. I also added some extra information.

CRD updated

enhancements/platform/agreement/README.md
JoseSzycho: Can the DocumentAcceptance be stored in the organization namespace?

By doing so, I could create an IAM policy binding that allows certain roles to list all DocumentAcceptance objects. 
These roles would then be able to retrieve specific DocumentRevision objects, since all authenticated users should have access to DocumentRevision.

The downside is that if the organization is deleted, all DocumentAcceptance objects stored there would be lost.

Do you have any alternative ideas or suggestions for handling this?

enhancements/platform/agreement/README.md
scotwells: I don't think it's needed, but curious what @jacobsmith928 / @zsmithnyc / @bmertens-datum might think.

enhancements/platform/agreement/README.md
JoseSzycho: Great. Will all agreements be associated with Organizations? I’m curious how users will be able to retrieve their documents.

Or will this be achievable after the feature that allows users to get the resources they have access to is implemented?

scotwells (approved):

PR #423, opened 2025-10-23
feat: init auth dns

zachsmith1:
- Introduce DNS Operator design under Design Details
- Add CRDs: DNSZoneClass, DNSZone, DNSRecordSet with Go struct snippets
- Describe controller responsibilities: upstream replicators and downstream reconcilers
- Document object mapping and condition semantics (Accepted, Programmed)
- Provide example YAML for DNSZoneClass, DNSZone, DNSRecordSet
- Outline future work: DNSProvider/Pool, NS strategies, admission webhooks
- Update Summary to emphasize provider-agnostic APIs and desired vs realized state

### DIFF
--- a/enhancements/authoritative-dns/README.md
+++ b/enhancements/authoritative-dns/README.md
@@ -260,7 +260,6 @@ type DNSZone struct {
 type DNSZoneSpec struct {
   DomainName string `json:"domainName"`
   DNSZoneClassName  string `json:"dnsZoneClassName,omitempty"`
-  AutoRegister bool `json:"autoRegister,omitempty"`
 }
 
 type DNSZoneStatus struct {


enhancements/authoritative-dns/README.md
```diff
@@ -0,0 +1,780 @@
...
+}
+
+type DNSZoneSpec struct {
+  DomainName string `json:"domainName"`
+  ClassName  string `json:"className,omitempty"`
+  AutoRegister bool `json:"autoRegister,omitempty"`
+}
+
+type DNSZoneStatus struct {
+  ZoneCreated       bool     `json:"zoneCreated,omitempty"`
```
joshlreese: Is this something that could be communicated via a condition?

I think we should have an `Accepted` and `Programmed` condition by default. These would align with conditions that are on HTTPProxies, [Gateway resources](https://gateway-api.sigs.k8s.io/geps/gep-1364/?h=conditions#new-and-updated-conditions), and other compute related resources. 

You can use a kubebuilder marker as seen [here](https://github.com/datum-cloud/network-services-operator/blob/b7c879b6ccdd310ce48ab922870f94c2e932a9fb/api/v1alpha/httpproxy_types.go#L228) to have the conditions automatically injected upon creation.

enhancements/authoritative-dns/README.md
```diff
@@ -0,0 +1,780 @@
...
+  Status DNSRecordSetStatus `json:"status,omitempty"`
+}
+
+type DNSRecordSetSpec struct {
+  ZoneName string      `json:"zoneName"`
+  Records  []DNSRecord `json:"records"`
+}
+
+type DNSRecord struct {
+  DNSName   string   `json:"dnsName,omitempty"`
```
joshlreese:
```suggestion
  Name   string   `json:"name,omitempty"`
```

enhancements/authoritative-dns/README.md
```diff
@@ -0,0 +1,780 @@
...
+  RecordType string  `json:"recordType"`
+  Targets   []string `json:"targets"`
+  RecordTTL *int64   `json:"recordTTL,omitempty"`
```
joshlreese: We should plan on adding a default TTL. Looks like CloudFlare's API defaults to `1`, which is "auto" - not sure what the details of that are, though.

```suggestion
  TTL *int64   `json:"ttl,omitempty"`
```

enhancements/authoritative-dns/README.md
```diff
@@ -0,0 +1,780 @@
...
+
+type DNSRecordSetSpec struct {
+  ZoneName string      `json:"zoneName"`
+  Records  []DNSRecord `json:"records"`
+}
+
+type DNSRecord struct {
+  DNSName   string   `json:"dnsName,omitempty"`
+  RecordType string  `json:"recordType"`
+  Targets   []string `json:"targets"`
```
joshlreese: We spoke about having a union type for records to support structured configuration. I think we should still plan on that.

That would mean we'd have fields for each record type that's supported, and each could have common fields as well as record specific fields. The [CloudFlare API docs](https://developers.cloudflare.com/api/resources/dns/subresources/records/methods/create/) show how various record types can have specialized fields (like the `data` fields in HTTPS and SRV record types).

Also, thoughts on using the term `content` and not `targets`, as target doesn't make sense for some records types like TXT?

A manifest would end up looking like this:

```yaml
spec:
  dnsZoneRef:
    name: example.com
  recordType: A
  records:
    - name: test
      a:
        content: 
          - 1.1.1.1
          - 8.8.8.8
    - name: test2
      a:
        content: 
          - 9.9.9.9
```

enhancements/authoritative-dns/README.md
```diff
@@ -0,0 +1,780 @@
...
+  RecordTTL *int64   `json:"recordTTL,omitempty"`
+  Labels    map[string]string `json:"labels,omitempty"`
+  ProviderSpecific []ProviderSpecificKeyValue `json:"providerSpecific,omitempty"`
```
joshlreese: I think we should omit this for now.

enhancements/authoritative-dns/README.md
```diff
@@ -0,0 +1,780 @@
...
+  metav1.TypeMeta   `json:",inline"`
+  metav1.ObjectMeta `json:"metadata,omitempty"`
+  Spec   DNSZoneSpec   `json:"spec"`
+  Status DNSZoneStatus `json:"status,omitempty"`
+}
+
+type DNSZoneSpec struct {
+  DomainName string `json:"domainName"`
+  ClassName  string `json:"className,omitempty"`
+  AutoRegister bool `json:"autoRegister,omitempty"`
```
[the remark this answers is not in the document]
joshlreese: Do you mean actual domain registration, or creating `Domain` resources?

enhancements/authoritative-dns/README.md
zachsmith1: will add this and also include a "raw" option as well

PR #551, opened 2026-01-13
chore: update TSigKey with DNSZoneTsigKey

zachsmith1:
Update the zone transfers enhancement to replace the prior `TSIGKey` proposal with a zone-scoped `DNSZoneTsigKey` resource and rename all dependent API surfaces accordingly. Document why keys are scoped per-zone initially (multi-tenant safety, simpler lifecycle/GC, reduced blast radius) while explicitly leaving room for a future shared/global TSIG key object. Clarify that `spec.keyName` is a TSIG Key Name per RFC 2845, is canonicalized to an FQDN (with optional trailing period), must match the zone domain (apex or subdomain), and that the backing Secret only contains the shared `secret` value. Update `ZoneTransfer` to reference `dnsZoneTsigKeyRef`, align validation/invariants and risk mitigation text around key-name isolation, add an Alternatives section showing a generic typed reference approach, and rename status to tsigKeyName to reflect the on-the-wire key name.


enhancements/authoritative-dns/zone-transfers/README.md
```diff
@@ -510,11 +530,48 @@ TBD
...
+    name: example-com
+  role: Secondary
+  secondary:
+    masters:
+      - address: 198.51.100.10
+    tsigKeyRef:
+      apiGroup: dns.datum.net
+      apiVersion: v1alpha1
+      kind: DNSZoneTsigKey
+      name: example-com-upstream
```
scotwells: I like the alternative approach here since it allows us to future-proof the API. I think you can drop the api group / api version though since we don't need to worry about supporting TSIGs from different API groups or versions.

PR #553, opened 2026-01-14
feat: milo project resource sets

zachsmith1:
Adds an enhancement/design doc proposing `ProjectResourceSet` / `ProjectResourceSetBinding` (inspired by Cluster API’s [ClusterResourceSet](https://github.com/kubernetes-sigs/cluster-api/blob/main/docs/proposals/20200220-cluster-resource-set.md#clusterresourcesetbinding-object-definition)) to declaratively apply Kubernetes manifests into each Project’s virtual control plane. The initial target is replacing the Project controller’s hard-coded default installers for `GatewayClass` and `DNSZoneClass`, with org-scoped policy via label selectors and clear status/traceability via bindings.


enhancements/platform/milo-projectresourcessets/README.md
```diff
@@ -0,0 +1,532 @@
...
+behavior into Project control planes at creation time (and optionally keep it up to
+date over time) without hard-coding per-service bootstrapping into the Project
+lifecycle controller.
+
+The motivating problems include:
+
+- **Separation of concerns**: the Project lifecycle controller should focus on
+  provisioning and readiness of Projects, not on installing service-specific resources.
+- **Consistency and policy**: Organizations often want consistent defaults across
+  all Projects they own.
```
scotwells: The wording here makes it sound like this resource is something that consumers would have access to in order to manage default resources created in their project. I'd expect this concept to be something that service providers are responsible for managing.

enhancements/platform/milo-projectresourcessets/README.md
```diff
@@ -0,0 +1,532 @@
...
+- **Idempotency**: application must be safe to retry. Prefer server-side apply (SSA)
+  for patch-like behavior and stable “field manager” ownership.
+
+### Risks and Mitigations
+
+- **Accidental wide blast radius** (e.g. selector matches too many Projects):
+  - Mitigation: require non-empty selectors; optionally gate with feature flags and
+    RBAC so only privileged actors can create/modify sets.
+  - Mitigation: record all applications in `ProjectResourceSetBinding` for audit and
+    rollback planning.
```
scotwells: I'm not understanding the risk / mitigation here. What would the typical use-case look like if we required non-empty selectors? I'd think in almost all cases, a resource set will apply to all projects, not a subset.

PR #556, opened 2026-01-19
Add connector + Envoy design doc

zachsmith1:
Adds a new connector/Envoy design doc for datum-connect/datum-desktop, detailing the end-to-end tunnel flow, sidecar metadata contract, and NSO programming responsibilities. It documents three Envoy Gateway implementation options with diagrams and validation notes, and clarifies MVP scope as CONNECT-TCP with a path to CONNECT-UDP. Open questions include which Envoy Gateway option to standardize on and whether to define an initial ConnectorClass (its intention and how it should work).


enhancements/networking/connectors/connector-envoy-design/README.md
```diff
@@ -0,0 +1,421 @@
...
+#### 2) Control plane publishes discovery records
+**Responsible:** Control Plane Controller
+
+**Data:**
+- `Connector.status.connectionDetails`
+
+**Actions:**
+- Publish DNS TXT records (or equivalent discovery data), e.g.
+  `_iroh.<node-id>.<project>.datumconnect.net`.
+- (Do we want to include project in the url?)
```
Frando: Including the project will mean that you can't resolve an iroh endpoint just by knowing its endpoint id. So I suggest to not include `<project>`. If we remove `<project>`, iroh can just use the regular DNS discovery to resolve endpoints, which will keep things simpler without any loss IMO.

Also, IIRC devices can publish proxies for multiple projects (maybe not in the MVP but in the future) so then we'd need multiple DNS entries for the same endpoint, which sounds cumbrsome to me.

zachsmith1: We wont be including the project here. I'm cleaning that up now

enhancements/networking/connectors/connector-envoy-design/README.md
```diff
@@ -0,0 +1,421 @@
...
+### Envoy + Sidecar Data Path
+
+1) Envoy receives a request from the client.
+2) Envoy routes the request to a tunnel listener (not the real backend).
+3) The tunnel listener uses `tcp_proxy` with `tunneling_config` to emit a CONNECT
+   request to the sidecar backend.
+4) The tunneling config injects headers describing:
+   - Connector NodeID (iroh public key)
+   - Destination host
+   - Destination port
```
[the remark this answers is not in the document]
zachsmith1: correct and good point. I'm removing those fields and same for the next comment on destination protocol

PR #625, opened 2026-03-01
feat(compute): add Functions serverless enhancement

scotwells:
## Summary

This enhancement introduces **Functions**, a serverless compute capability that makes it easy for developers to deploy code without managing infrastructure.

> [!NOTE]
>
> I'd recommend reading the [rendered view of the markdown document](https://github.com/datum-cloud/enhancements/blob/feat/functions-enhancement/enhancements/compute/functions/README.md). 

## What developers get

- **Deploy in seconds**: Push a container image and your function is live—no network config, instance sizing, or placement decisions
- **Pay only for what you use**: Functions scale to zero when idle and spin up instantly when traffic arrives
- **Fast response times**: Sub-50ms cold starts mean users don't wait for your function to wake up
- **Use your favorite language**: Go, Node.js, Python, and Rust supported out of the box

## Common use cases

- API endpoints and webhooks
- Request interception (auth, header modification, URL rewriting)
- Server-side rendering at the edge

## Architecture highlights

Functions builds on the existing Workload primitive, so it inherits all the platform's battle-tested compute infrastructure. Private connectivity via Service Connect means traffic between your apps and functions never touches the public internet.

## What's included

- Enhancement README with motivation, goals, and design rationale
- Three C4 architecture diagrams showing system context, control plane, and connectivity
- Shared diagram theme for consistent styling


scotwells:
I'm also thinking that we reduce the scope of this to only support Javascript / Python for the MVP. 

Golang / Rust will require the platform to compile code into executables before running. That's a more complex pipeline with more things that can go wrong, more error messages to surface clearly, and more edge cases to handle.

jacobsmith928 (commented):
LGTM, thanks for your deep dive on this @scotwells

jacobsmith928 (approved):
I meant to approve on my last review

drewr (approved):
Thanks for this @scotwells. I didn't see anything that stood out to me as a concern at this stage. I like the control plane separation. Looking forward to diving more into how auth will work.

jacobsmith928:
@scotwells yes that seems like a good place to start vis a vis:
https://developers.cloudflare.com/workers/languages/

scotwells:
> What are your thoughts on extending the dev experience (maybe future) to support a use-case around auto shipping/building/publishing with something like buildpacks? (https://buildpacks.io/) It would give a magical "push this repo to a cloud function" feeling.

@zachsmith1 the **_Git-driven approach_** I had in the **Developer Experience** lightly touches on this. Do you think it's worth expanding? I suspect this is something we'll tackle in a second phase since it requires we have a build pipeline to pull artifacts from GitHub and build them into an OCI image we can use for the function.

> It would be helpful if we could go into more detail around the update/upgrade functionality for functions. If the user wanted to traffic split across two functions easily, how would we support that? I'd think the same process would happen during an upgrade.
> 
> Ideally, we can support something like:
> 
> - if I want to upgrade my function to my latest app version, how does datum provide 0 downtime upgrades.
> - if the new image isn't healthy, how can we assure the user we will roll back to a version that works.
> - if I want to spin up a function version-B, how can I split traffic or maybe shadow requests to it (future)

Good suggestion. I'll need to think about how to model this. It's is a little more involved since we want "functions" to remain generic to what triggers it. 

This may need to be a higher-level deployment orchestration primitive since it would need to deploy the new function version and then update the gateway to do[ traffic splitting](https://gateway.envoyproxy.io/docs/tasks/traffic/http-traffic-splitting/#weighted-backendrefs) and monitor for changes in response codes.

bmertens-datum (commented):
Hello @scotwells, 
Looks good. 
I don't want to block this from moving forward. I would like to have a chat about "Request interception (auth, header modification, URL rewriting)"  This can be done much faster with LUA.  I'll put a meeting in the calendar.

bmertens-datum (approved):
Hello @scotwells, 
Looks good. 
I don't want to block this from moving forward. I would like to have a chat about "Request interception (auth, header modification, URL rewriting)"  This can be done much faster with LUA.  I'll put a meeting in the calendar.

zachsmith1 (approved):

PR #646, opened 2026-03-06
Workload Identity Federation for External Platform Authentication

scotwells:
## Summary

Workload Identity Federation lets external platforms — GitHub Actions, GCP, AWS, and Azure — authenticate to Milo using their native OIDC tokens instead of long-lived credentials.

### The Problem

Today, teams that want CI/CD pipelines or cloud workloads to access their Milo projects must:
1. Create MachineAccounts with long-lived keys
2. Store those keys as secrets in GitHub, cloud parameter stores, etc.
3. Manually rotate and revoke credentials
4. Hope nothing leaks

This creates credential management burden, security risk from stored secrets, and audit difficulty.

### The Solution

With workload identity federation, external platforms authenticate using their **native short-lived OIDC tokens**:
- A GitHub Actions workflow uses its job token
- A GCP Cloud Function uses its service identity token
- No credentials are shared or stored

### How It Works

- **Trusted issuers, ready out of the box.** The platform pre-registers OIDC issuers for GitHub Actions, GCP, AWS, and Azure. You reference one — you don't configure issuer URLs or JWKS yourself.
- **Bring your own issuer, too.** Have an identity provider the platform hasn't pre-vetted — a private Kubernetes cluster, SPIRE, Okta? Register it directly on your own issuer instead of waiting on a platform team.
- **Rules, not resources per access level.** Want to restrict deploys to `main` while allowing read-only access from any branch? That's two lightweight rules against the same issuer, not two copies of your OIDC config. Each rule uses a CEL expression over token claims (repository, branch, actor, etc.) to decide what it matches.
- **One kill switch per issuer.** Disable an issuer and every rule under it stops authenticating immediately — no need to track down and disable rules one at a time.
- **Bind directly in PolicyBinding.** Reference a rule by name as a PolicyBinding subject — the platform resolves the underlying identity for you, so there's no URI to hand-copy or typo.

### Verifying Your Setup

Before a real pipeline depends on it:
- **Verify issuer** dry-runs your JWKS configuration and flags reachability problems immediately.
- **Test rule** opens a live 15-minute window — trigger a real token exchange from your workload and watch it succeed or fail in real time, with the specific reason (bad signature, audience mismatch, condition not met) if it doesn't.
- **Authentication history** logs every authentication attempt against a rule, successful or not — so you can tell "this workflow never tried" from "it tried and was rejected," and catch a pipeline under sustained failed attempts before it becomes an incident.

### Supported Platforms (v1)

- GitHub Actions
- GCP (Cloud Functions, Cloud Run, etc.)
- AWS (Lambda, ECS, etc.)
- Azure AD

### Security Benefits

- **No stored credentials**: OIDC tokens are short-lived and auto-provisioned
- **Immediate revocation**: Disable a rule or an issuer — or delete the PolicyBinding — for instant access removal
- **Full audit trail**: Every authentication attempt is logged, successful or not, with the matched claims
- **Attribute-based control**: Restrict by repository, branch, environment, or any OIDC claim
- **Fewer places to make a mistake**: pre-vetted issuers, direct rule references in PolicyBinding, and pre-flight verification catch misconfiguration before it reaches a real workflow

---

📄 See the full enhancement document for design details, user stories, and resource reference.

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/enhancements/platform/identity-and-access-management/README.md
+++ b/enhancements/platform/identity-and-access-management/README.md
@@ -195,3 +195,5 @@ In the future we will also support the following formats to support additional
 use-cases.
 
 - `group:<group-email>` grants a group in the IAM system access to a role
+- `principal:<principal-uri>` grants a federated identity access to a role (see
+  [Workload Identity Federation](./workload-identity-federation/))
--- a/enhancements/platform/identity-and-access-management/workload-identity-federation/README.md
+++ b/enhancements/platform/identity-and-access-management/workload-identity-federation/README.md
@@ -1,0 +1,491 @@
+---
+status: provisional
+stage: alpha
+latest-milestone: "v0.1"
+---
+
+<!-- omit from toc -->
+# Workload Identity Federation
+
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [How It Works](#how-it-works)
+  - [User Stories](#user-stories)
+  - [Supported Platforms](#supported-platforms)
+  - [Security](#security)
+  - [Notes/Constraints/Caveats](#notesconstraintscaveats)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [Resource Model](#resource-model)
+  - [Authentication Flow](#authentication-flow)
+  - [Authorization Integration](#authorization-integration)
+  - [Attribute-Based Access Control](#attribute-based-access-control)
+- [Future Work](#future-work)
+- [Dependencies](#dependencies)
+- [Alternatives](#alternatives)
+
+## Summary
+
+Workload Identity Federation enables external platforms (GitHub Actions, GCP,
+AWS, Azure) to authenticate to Milo using their native OIDC tokens instead of
+long-lived credentials. Project owners configure trust relationships through
+WorkloadIdentityPool and WorkloadIdentityPoolProvider resources, and federated
+identities become subjects in PolicyBindings using principal URI strings.
+
+This approach eliminates credential management burden, removes the need to store
+secrets in CI/CD systems, and provides automatic token expiration with clear
+audit attribution.
+
+## Motivation
+
+Many teams need external platforms to access their Milo projects for CI/CD
+pipelines, automation, and third-party integrations. Common scenarios include:
+
+- **GitHub Actions workflows** deploying resources to a project
+- **Cloud platform workloads** (GCP Cloud Functions, AWS Lambda) calling Milo
+  APIs
+- **Third-party services** like Grafana reading project metrics
+
+Today, these use cases require creating MachineAccounts with long-lived
+MachineAccountKeys. This approach has significant problems:
+
+- **Credential management burden**: Teams must generate, store, rotate, and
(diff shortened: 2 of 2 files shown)


kprabhak:
@scotwells Written up a more detailed analysis of different [use cases for workload identity](https://github.com/datum-cloud/enhancements/blob/iam/scenarios-and-prds/enhancements/platform/identity-and-access-management/authentication/IAM-architecture-and-use-cases.md#2-use-cases) as well as a [draft PRD](https://github.com/datum-cloud/enhancements/blob/iam/scenarios-and-prds/enhancements/platform/identity-and-access-management/authentication/W1-workload-identity-federation-PRD.md) further tweaking this proposal. Please take a quick glance, and lets discuss further.

scotwells:
@kprabhak is there a PR where I can provide feedback on these documents? The first one seems to be trying to cover too much of the system instead of breaking the IAM system down into its different functional groups (e.g. Authentication, Authorization, etc). Creating more focused architecture diagrams will help separate concerns between different parts of the architecture. 

On the workload identity architecture, I would not expect the milo apiserver to trust or receive any tokens from external systems. Clients using workload identity federation will need to be able to exchange their token for an auth token provisioned by Datum's Identity service (e.g. Zitadel) so it's trusted by the platform. The milo apiserver should only ever trust tokens minted by Datum's identity service.

kprabhak:
> @kprabhak is there a PR where I can provide feedback on these documents? The first one seems to be trying to cover too much of the system instead of breaking the IAM system down into its different functional groups (e.g. Authentication, Authorization, etc). Creating more focused architecture diagrams will help separate concerns between different parts of the architecture.

Fair. I have separate docs for AuthN vs AuthZ, but tried to simplify by combining them. Let me clean up those, and create a separate PR for the federate-in PRD than for the arch/use-cases doc. 


> 
> On the workload identity architecture, I would not expect the milo apiserver to trust or receive any tokens from external systems. Clients using workload identity federation will need to be able to exchange their token for an auth token provisioned by Datum's Identity service (e.g. Zitadel) so it's trusted by the platform. The milo apiserver should only ever trust tokens minted by Datum's identity service.

That would have been my preferred approach too. Unfortunately, a [Zitadel limitation](https://github.com/datum-cloud/enhancements/blob/iam/scenarios-and-prds/enhancements/platform/identity-and-access-management/authentication/W1-workload-identity-federation-PRD.md#43-option-3--token-exchange-at-the-identity-provider--eliminated-by-test) prevents token exchange for tokens with foreign issuers. So our alternative options might be:

1. Collab with zitadel on a fix (including potentially doing this ourselves & contributing). Let me poke around to see the scope of what this might involve.
2. Consider a [STS component](https://github.com/datum-cloud/enhancements/blob/iam/scenarios-and-prds/enhancements/platform/identity-and-access-management/authentication/W1-workload-identity-federation-PRD.md#45-option-5--a-datum-side-security-token-service) in front of zitadel (such as a sidecar). Not as complex as it sounds, I've built these before for agent delegation scenarios, however it does mean another component to manage. It might also be useful for the W4 and W7 use cases (if/when we get to Edge/Tunnel auth, and agent delegation scenarios).

scotwells:
@kprabhak having a separate component handle the token exchange makes sense. Starts us down the journey of removing our reliance on Zitadel for our auth stack.

Issue #505, opened 2025-12-15
Fraud and abuse - Initial implementation


drewr:
## Problem

One of the main benefits of a cloud platform is that anyone can sign up and start consuming resources. This motion is the reason cloud has become the dominant way to consume infrastructure -- there isn't an IT department in the way.

However, this freedom is often abused by a user who signs up with a stolen credit card, disputes legitimate charges, abuses freemium services, or otherwise utilizes a company's resources outside their intended and sanctioned use.

At minimum, we need to isolate these types of users so that we can fully verify that they are malicious and then take action. There are two major parts to the problem:

1. Determine whether a user is malicious
2. Prevent the user from consuming resources

## Requirements

For the second part, we already have a deactivation procedure for users. This needs to make sure that any resources they've managed to use are terminated when the user is deactivated. We should then note on their account why (using our new Note resources from #473).

For the first part, the Datum team will ultimately decide how to handle a user, but the system can take steps on behalf of the team (Milo administrator), including but not limited to:

* Reach out to a third-party fraud detection service
* Check community-maintained denylists for known-bad email addresses or domains, IP ranges, embargoed origin countries (OFAC), etc. (If that doesn't exist, we can start a Milo version of it.)
* Other local Datum policies and environmental clues we discover

At a high level, this probably means some kind of Milo interface that generates a user fraud score and interoperates with other tools that we and our partners would likely work with.

## Next steps

- [ ] Research options
- [ ] Formulate plan for what can be done in ~1 week
- [ ] Get feedback here
- [ ] Implement initial functionality, adjust plan
- [ ] Deploy in "observe" mode (like we did with WAF) to production
- [ ] Watch for effectiveness and scoring decisions
- [ ] Sync on next steps after

scotwells:
@drewr overall agree on the scope of the issue. 

In a later phase, we could have this system to periodically analyze a user's activity from [the new activity API we're building](https://github.com/datum-cloud/engineering/issues/90) and determine if there's suspicious or fraudulent behavior.

> For the second part, we already have a deactivation procedure for users. This needs to make sure that any resources they've managed to use are terminated when the user is deactivated. We should then note on their account why (using our new Note resources from https://github.com/datum-cloud/enhancements/issues/473).

I'd consider using the `Reason` / `Description` fields on the UserDeactivation resource (see [API docs](https://github.com/datum-cloud/milo/blob/main/docs/api/iam.md#userdeactivationspec)) to indicate why a user was deactivated.

[...]

PR #661, opened 2026-03-11
feat: fraud and abuse prevention api

zachsmith1:
Introduces the enhancement proposal for a Kubernetes-native fraud and abuse prevention API (fraud.miloapis.com/v1alpha1). 

The system defines three core CRDs -- `FraudProvider`, `FraudPolicy`, and `FraudEvaluation` -- that together form a pluggable, staged pipeline for evaluating user fraud risk. 

- Providers are configured independently and orchestrated by a policy that supports layered execution with short-circuit logic (cheap checks gate expensive ones) and required stages (e.g. sanctions screening always runs). 

- v1 targets two built-in providers (MaxMind minFraud for IP/email risk scoring and moov-io/watchman for OFAC sanctions screening), deploys in `OBSERVE` mode to tune thresholds before enabling enforcement, and integrates with the existing Milo User and UserDeactivation lifecycle via the auth-provider-zitadel actions server. 

Designed to evolve toward intra-stage DAG execution, scheduled re-evaluation, data-change triggers, and webhook-based custom providers in future iterations.

Ref: https://github.com/datum-cloud/enhancements/issues/505

### DIFF
--- a/enhancements/platform/fraud-and-abuse/README.md
+++ b/enhancements/platform/fraud-and-abuse/README.md
@@ -1,0 +1,1314 @@
+---
+status: provisional
+stage: alpha
+latest-milestone: "v0.0"
+---
+
+# Fraud and Abuse Prevention API
+
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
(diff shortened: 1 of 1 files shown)


drewr:
Great that we have an API, but that's only one piece of handling fraud and abuse. We need a bigger picture of the plan:

* What's the user experience?
* When does the API get called?
* What happens if a user is deemed fraudulent? Who gets notified?
* How do we, as an organization, handle failure scenarios?

drewr (approved):
Thanks for the explanation!

Issue #577, opened 2026-01-22
Feature-flagging design


drewr:
### Research Goals

Today all users on Datum Cloud get the same view of the world. If we want to try out a feature on a small group of beta-testers we have to inflict that feature on everyone. We need an ergonomic way to map users to capabilities.

There is a Milo concept of entitlements that pairs with quotas that we likely want to work towards. Unsure here whether we can make incremental progress on something good enough or have to bite off the whole feature.

### Deliverables

2-3 options with tradeoffs.

### Additional Context

Similar functionality to Posthog, LaunchDarkly, etc.

kevwilliams:
https://github.com/datum-cloud/enhancements/pull/695 is the PR where I have put the proposal to use the quota system. I have also started to break out the implementation tasks into issues you can see in the milo repo starting with https://github.com/datum-cloud/milo/issues/575

PR #695, opened 2026-04-17
feat: feature-flagging via Milo entitlements (#577)

kevwilliams:
## Summary

Adds a provisional enhancement for org-level boolean feature flags built entirely on the existing `quota.miloapis.com/v1alpha1` primitives.

## What this proposes

- **Feature flag = `ResourceRegistration`** with `type=Entity`
- **Grant = `ResourceGrant`** with `amount=1` targeting an org
- **Check = `AllowanceBucket.status.available > 0`** queried by field selector
- v1 is zero-code: no new CRDs, no new controllers, no binary changes

## Key decisions needed from reviewers

The doc marks six open questions with `<<[UNRESOLVED]>>` blocks. The two blocking ones:

1. **`claimingResources` validation** — the CRD schema may enforce `MinItems=1`, which would break the proposed `ResourceRegistration` shape. Needs verification against the live CRD before this can move to `implementable`.
2. **Consumer check strategy** — confirm `AllowanceBucket` (Option A) as the standard read path, or prefer querying `ResourceGrant` directly (Option B).

## Test plan

- [ ] Verify `claimingResources: []` is accepted by the live `ResourceRegistration` CRD
- [ ] Apply sample `ResourceRegistration` and `ResourceGrant` YAML to a dev cluster and confirm `AllowanceBucket` is auto-created
- [ ] Confirm `AllowanceBucket.status.available` drops to 0 after `ResourceGrant` deletion
- [ ] Confirm org admin can list `AllowanceBuckets` with existing `organization-quota-manager` role

### DIFF
--- a/enhancements/platform/feature-flagging/README.md
+++ b/enhancements/platform/feature-flagging/README.md
@@ -35,10 +35,13 @@ organizations before a full rollout. This enhancement defines an org-level
 on the existing `quota.miloapis.com/v1alpha1` entitlement primitives — no new
 CRD kinds, no binary changes, no third-party dependencies.
 
-A feature flag is a `ResourceRegistration` with `type=Entity`. Granting a flag
-to an organization is a `ResourceGrant` with `amount=1`. Services check
-whether the flag is enabled by querying the auto-maintained `AllowanceBucket`
-for that (org, resourceType) pair.
+A feature flag is a `ResourceRegistration` with `type=Feature` (requires a
+one-line Milo CRD prerequisite; see [Incremental Path](#incremental-path)).
+Granting a flag to an organization is a `ResourceGrant` with `amount=1`.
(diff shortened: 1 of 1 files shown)


scotwells:
@kevwilliams the registration path looks okay, the type being `Feature` might make more sense. Maybe we can look at something where we can provide a check API that's compatible with https://openfeature.dev so we can leverage its client-side tooling for feature checks? Not opposed to just building a client SDK for feature checking against the quota service either if we want to roll our own.

kevwilliams:
@scotwells Updated it to include these things, also considered the rolling our own idea a bit, I think OpenFeature gives us quite a bit for free there so I included that in there too.

scotwells (dismissed):

kevwilliams:
Closing because of inactivity, if we need it, we can open something again.

PR #705, opened 2026-05-01
feat: BGP control plane architecture

privateip:
This commit provides the first draft of the proposed BGP control plane for Galactic VPC.  It includes the recommendation for a tiered route reflector hierarchy and defines how worker nodes will connect to the infrastructure.


slindseysr:
Do you have a concept diagram of what the physical and logical data plane topology looks like?  I am trying to wrap my head around how the POP routers resolve nexthop of the iBGP routes they learn from the RRs.

privateip:
I'm heads down focused on building out the gvpc topology in containerlab today.  Once we have that committed we can start to look at the real routing tables and configuration as opposed to synthesized models and topology diagrams.   I expect to have a PR up today yet that will be the two tier, three region topology.

slindseysr:
Per our zoom conversation, the design assumes global reachability in the underlay already in place.  Details on requirements documented in #709

privateip:
> design assumes global reachability in the underlay already in place

correct

PR #712, opened 2026-05-04
feat: Galatic VPC underlay network requirements

privateip:
This enhancment defines the requiremnts that must be supported by **any** underlay network in order to successfully deploy and operate Galactic VPC over the top.

### DIFF
--- a/enhancements/networking/underlay-network/README.md
+++ b/enhancements/networking/underlay-network/README.md
@@ -1,0 +1,283 @@
+# Galactic VPC — Underlay Network Requirements
+
+**Reference design:** [BGP Control Plane Architecture — PR #705](https://github.com/datum-cloud/enhancements/pull/705)
+
+---
+
+## Purpose
+
+This document defines the underlay network requirements that the BGP control plane design (PR #705) implicitly depends on but does not specify. That design makes a set of hard assumptions about the physical and logical substrate beneath it. Those assumptions are load-bearing: if the underlay fails to satisfy them, the overlay breaks in ways that are difficult to debug and harder to fix post-deployment.
+
+This is a requirements document, not a deployment guide. It defines *what* the underlay must provide, not *how* to provide it.
+
+---
+
+## Foundational Constraints
+
+These are non-negotiable. Every requirement below flows from them.
+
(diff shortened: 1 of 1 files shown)


slindseysr:
Nice detail here @privateip!  Some thoughts below.

- The MTU requirement of >=9080 is achievable for a private backbone.  However, there could be some scenario where POP(s) only have Internet transit and thus maximum MTU of 1500.  If we intend to extend inner MTU of 1500 then we will need to support inline fragmentation and reassembly.  We should consider enabling MSS clamping to 1400 for these cases unless we have confidence that our hardware can handle the fragmentation with negligible impact on latency.
- Backbone underlay BFD timers should be set based on a calculation that includes a factor of latency of the link they are protecting and tuned to desired detection and failover targets. Longer links will have longer timers but detection and reroute should easily be far below the stated minimum to hold BGP up.  The Internet path overlay will be a bigger challenge but all failure detection and failover mechanisms available should be implemented. 
- Infrastructure prefixes should be filtered as described with exception of the router loopbacks.  These must be part of a block advertised publicly for interop with transit only sites and/or failover to Internet based paths.  The routers should be protected with an aggressive CoPP policy.

privateip:
> The MTU requirement of >=9080 ....

Good point, I will made some updates to the document to address this

> Backbone underlay BFD timers should .....

I don't disagree with your comment here.  I have tried to keep this document as a set of requirements meaning non implementation specific.   That said, I can add some additional language to address this.

> Infrastructure prefixes should be filtered ....

You are absolutely correct here, just a miss on my part

privateip:
@slindseysr I have updated the document per your comments.  Please review

slindseysr (approved):
The updates here look good, and good call adding the extra 40 bytes of headroom. Since the comments earlier, we have had internal discussion around MTU and whether we even want to offer gVPC at sites that cannot support MTU >1500. There are some tradeoffs we need to evaluate further, but operationally we may decide to only offer gVPC at sites that can support the full scale MTU. If we go this route, this should be straightforward to implement by pruning BGP advertisements for Anycast edge-service prefixes at Internet only sites. Those Internet only PoPs would still handle DNS Anycast and any other services that do not require backhauling. Regardless what we ultimately decide, I think we should still fully test gVPC from the Internet only PoPs and validate our interop solution.

PR #740, opened 2026-05-23
feat: Datum Cloud IPv6 addressing plan

privateip:
Summary

Refactors the IP Addressing Plan into a uSID-first architecture aligned with RFC 9374, and splits the SRv6 SID specification into its own document.

What changed

Moved enhancements/networking/addressing-plan/README.md → enhancements/networking/ipv6/README.md

The addressing plan has been rewritten to reflect a micro-SID (uSID) deployment:

- uSID Block terminology — The per-PoP /48 locator is now described as a uSID Block (uFMT 48+16 carrier format). All nodes within a PoP share the same /48 prefix; traffic steering is handled by the 16-bit Active uSID slot (bits 49–64) rather than per-router locator subnets.
- Internal infrastructure addressing — Added a dedicated /52 block carved from the PoP ULA /48 for platform-owned subnets (compute node, management, reserved), separate from the customer VPC pool.
- uSID VRF scaling limit — Documented the hard architectural boundary of 65,535 concurrent VRF IDs per PoP domain (16-bit argument field), with exhaustion procedures for when a PoP reaches 80% utilization.
- Expanded hierarchy table — The allocation summary now includes Per-Node Instruction + Behavior (Active uSID) and Per-Tenant VRF Context (Argument) rows to make the SID bit layout explicit.
- uSID-aware security properties — The isolation section now describes how the uSID Block, Active uSID slot, and VRF argument are opaque to underlay routing and do not affect prefix filtering.
- RIR-generic language — Replaced ARIN-specific references with generic RIR PI terminology to support future multi-region procurement.
- RIR reference table — Added a reference table of the five RIRs and their regions for operational clarity.
- DFZ callout — Moved the DFZ definition into a callout note in the pools section for better discoverability.
- GUA Tenant Addressing Policy — Retained in full (default deny export, explicit allowlist, RPKI/ROA requirements, BYOP validation, route leak detection, emergency quarantine).

Added enhancements/networking/srv6/README.md

A new dedicated document for the SRv6 uSID plan that was previously embedded in the addressing plan:

- uSID carrier layout — uFMT 48+16 compressed structure with uSID Block, Active uSID, Next uSID/Argument, and Padding fields.
- Behavior registry — uEnd.DT4, uEnd.DT6, uEnd.DT46 with platform-wide code assignments and a warning about uEnd.DT46 hardware pipeline requirements.
- VRF ID argument semantics — PoP-local 16-bit VRF ID allocation, infrastructure mapping steps, and relationship to the VRF instance.
- SIDs per tenant VRF — One active uSID sequence per behavior, with a table covering IPv4-only, IPv6-only, dual-stack single-table, and dual-stack per-AF configurations.
- VRF ID space and scale — 65,535 usable VRF IDs per PoP node, 80% utilization alerting threshold, and elimination of secondary locator block steering during scaling.

Deleted enhancements/networking/addressing-plan/README.md

Superseded by the new ipv6/README.md + srv6/README.md split.

Cross-references

- IP Addressing Plan — address pool allocations, VPC options, capacity, isolation
- SRv6 uSID Plan — SID structure, behavior codes, VRF ID semantics, scaling
- Underlay Network Requirements — BGP peering, loopback reachability, backbone link addressing


scotwells (dismissed):

privateip:
@slindseysr can you review this PR for accuracy and alignment?

privateip:
@0xmc please review the suggested IPv6 addressing plan and either approve or provide feedback

0xmc:
@privateip feedback from Claude which I have reviewed/edited.  In three parts.  This is the first one.

# Design question: should the tenant pool be carved per-PoP, or flat across PoPs?

The locator /48 and the infra loopback /48 both have clear per-PoP justifications — locators are advertised per-PoP into the underlay, and infra loopbacks aggregate per-PoP for external advertisement. The tenant /48 doesn't share those constraints: ULA space is never advertised externally, and SRv6 doesn't need the inner tenant address to be PoP-aggregatable — the locator does that work in the underlay routing table.

With the tenant pool carved per-PoP, a VPC present at multiple PoPs ends up with disjoint /64s from different /48s — no single "VPC CIDR" the tenant can point to, and the PoP is recoverable from the prefix (which the doc's "topology not in address bits" line suggests we'd rather avoid).

A flat tenant pool would be one alternative: one CIDR per VPC, valid at every PoP the VPC exists in, with IPAM tracking (VPC ↔ PoPs) by association. Each PoP still holds a VRF for the VPC; the per-PoP routing scope lives in the SRv6 SID rather than in the tenant prefix.

Both shapes are defensible. Let's produce an artifact of why we're choosing one over the other.

Related sizing question: a /64 per VPC gives the tenant a single subnet with no room to subdivide. If the platform is intentionally subnet-less, worth saying so explicitly; if not, a larger per-VPC allocation (tier-dependent) is closer to what tenants will expect from AWS/GCP/Azure semantics.

0xmc:
Second one

# Smaller clarity item: SID bit budget

The SID layout adds to 48 + 32 + 16 = 96 bits, leaving 32 unaccounted out of 128. Could be worth showing the full bit layout — whether the remainder is a node-ID, padding, or used differently — so the structure is self-explanatory at a glance.

0xmc:
Third and final

# ULA vs GUA

## Hygiene: RFC 4193 specifics for the ULA pool

RFC 4193 actually defines ULA as fc00::/7 — a /7, with bit 7 (the "L bit") splitting it into two halves:

  - fd00::/8 (L=1): locally assigned, use freely
  - fc00::/8 (L=0): reserved for a future centrally-assigned ULA registry

That central-assignment registry was proposed in the mid-2000s and never came to exist. fc00::/8 is therefore unused in practice — but it's also not assigned to anyone, and the spec specifically says local assignments belong in fd00::/8.

There's a second piece of the spec that matters here: the next 40 bits after the prefix are meant to be a pseudo-randomly generated Global ID (RFC 4193 §3.2.2). The Global ID's job is to give every ULA-using network a probabilistically unique /48, so that two such networks can later be merged, peered, or VPN'd together with a ~2^-40 collision rate per /48. Without a random Global ID, ULA effectively becomes "the well-known prefix everyone uses" — operationally about the same as 10.0.0.0/8.

The current doc treats fc00::/8 as a flat pool to carve from, which sidesteps both pieces. Moving to fd00::/8 plus a random 40-bit Global ID (one per platform, or one per region if regional aggregation is wanted later) brings the plan into conformance without altering anything structural.

## Where this actually bites in practice

A few concrete scenarios where collision-free ULA matters:

  - Tenant site-to-site / VPN bringup. A tenant connects on-prem to a Datum VPC. They almost certainly already use ULA on-prem — every Mikrotik, OPNsense, and Ubiquiti deployment ships with example configs in fd00::/8. If their on-prem ULA collides with the Datum-assigned /48, the fix is renumbering one side.
  - Multi-cloud overlays. Tenant runs Datum + AWS Transit Gateway / GCP Cloud Interconnect / Azure ExpressRoute and stitches them. Other clouds either use GUA (no collision) or let the tenant bring ULA (possible collision). Whoever has the random Global ID wins the renumbering argument.
  - M&A on Datum's side. If Datum ever acquires or peers with another operator running a similar overlay, both having "fc00::/8 raw" guarantees collision across the whole space.
  - RFC 6724 source-address selection. On a dual-stack host with both GUA and ULA addresses, ULA is preferred below IPv4 in the default policy table. Tenants
  who get a ULA-only assignment and expect IPv6-preferred behavior will be surprised.

## The bigger question: should this even be ULA?

AWS, GCP, and Azure all use GUA (their own ARIN/RIPE allocations) for tenant VPC addressing, even when traffic stays internal to the cloud. The rationale:

  - GUA is guaranteed unique, so all the collision scenarios above disappear by construction.
  - Tenants can connect overlay-internal addresses to external services without NAT or renumbering pain.
  - Filtering still works — "don't advertise this prefix to the DFZ" is a one-line route policy whether the prefix is GUA or ULA.
  - Costs one more ARIN allocation, which is cheap.

The argument for ULA in Datum's case is the SRv6 model: tenant packets are always encapsulated, so the inner address never sees the public Internet, and ULA is a stronger semantic statement of intent ("this is not Internet-routable, full stop"). That's defensible — but it's a deliberate trade between semantic clarity and collision risk, and the doc would benefit from saying which trade we're making and why, rather than picking ULA by default.

## Cleanest path forward

### If the design intent is to stay with ULA:

  - Switch the pool from fc00::/8 to fd00::/8
  - Generate a 40-bit Global ID per RFC 4193 §3.2.2 (one for the platform, or one per region) and record it in the same Milo IPAM entry as the placeholder prefixes
  - Add a sentence in the doc acknowledging the trade vs. GUA and why ULA was chosen

### If the design intent is to move to GUA:

  - Add the tenant GUA block as a fourth top-level pool (alongside the SRv6 locator and infra loopback blocks)
  - Apply the same external-advertisement filtering policy
  - No collision concern, no Global ID, no RFC-compliance debate

drewr:
Helpful info, thanks @0xmc. Can you summarize all that?

0xmc:
@drewr here's the summary:

- The proposal has a /48 per PoP for the tenant overlay VPC, but it the tenant has a VPC that spans multiple PoPs, their VPC will be comprised on non-aggregated /48s.  This differs from the typical default one-CIDR-per-VPC which can then be subnetted per region/AZ/PoP.  It is, however, possible I am missing a critical part of the design philosophy.
- Regardless, but related, to the above is do we want to use ULA as proposed or GUA?  GUA does provide guarantees for uniqueness and expected routing family selection preferences that ULA does not.  Regardless of ULA vs GUA, we should document why we're choosing one over the other.
- If we do decide to go the ULA path, we should use fd00::/8 as it may help avoid collisions in the cases of tenant-onboarding, M&A, or multi-cloud overlays.
- Unrelated to all of the above, the SID format is a little unusual with 96 bits instead of 128, and the 32-bit function field is bigger than the typical (or so I've read) 16 bits.

drewr:
Ok, sounds like there are a couple open questions there that are blocking. Probably better to add them inline in the doc so we can adjust the language more specifically?

enhancements/networking/addressing_plan/README.md
```diff
@@ -0,0 +1,156 @@
...
+```
+|<------- 48 bits ------->|<------- 32 bits ------->|<-- 16 bits -->|
+        Locator                    Function               Arguments
+   [LOCATOR-BLOCK]:X:Y::/48       e.g. End.DT4/VRF         VRF ID
+```
+
+| Field     | Bits | Description                                           |
+| --------- | ---- | ----------------------------------------------------- |
+| Locator   | 48   | Identifies the PoP; drawn from `[LOCATOR-BLOCK]/33`   |
+| Function  | 32   | Identifies the SRv6 behavior (e.g., End.DT4 per VRF)  |
```
0xmc: 32 seems high.  https://datatracker.ietf.org/doc/html/rfc8986#name-srv6-endpoint-behaviors-reg fits in 16.

enhancements/networking/addressing_plan/README.md
```diff
@@ -0,0 +1,156 @@
...
+
+Each PoP receives a `/48` from the platform's registered SRv6 locator block. This `/48` is the SID locator for all SRv6 SIDs originated at that PoP (e.g., `End.DT4`, `End.DT6`, `End.DT46` per-VRF SIDs). SIDs are drawn from this block.
+
+The locator block is PoP-scoped. All SIDs for a given PoP share the same `/48` locator prefix, enabling aggregate advertisement and clean per-PoP filtering.
+
+The PoP's `/48` locator prefix is advertised into the underlay via BGP IPv6 Unicast. This is not optional — underlay reachability of the locator is a prerequisite for SRv6 forwarding. Without it, remote PoPs cannot route encapsulated packets to their destination. The aggregate `/48` is advertised, not individual SIDs.
+
+### SRv6 SID Structure
+
+SIDs within a PoP's `/48` locator follow a fixed-length layout: 48-bit locator, 32-bit function, 16-bit argument.
```
0xmc: Should these add up to 128?

enhancements/networking/addressing_plan/README.md
```diff
@@ -0,0 +1,156 @@
...
+| SRv6 SID locators        | `[LOCATOR-BLOCK]/33` (ARIN PI) | SRv6 segment identifiers                      | Advertised in the underlay (BGP IPv6 Unicast); reachability is required for SRv6 forwarding |
+| Infrastructure loopbacks | `[INFRA-BLOCK]/33` (ARIN PI)   | Router loopbacks for underlay and overlay BGP | Underlay-reachable; aggregate-only to external peers                                        |
+
+> **Note on registered blocks:** Both the SRv6 locator block and the infrastructure loopback block must be Provider Independent (PI) space registered with ARIN. Using Provider Aggregatable (PA) space assigned by an upstream carrier is not acceptable — it creates a hard dependency on that carrier and prevents portability. The actual prefix values are placeholders in this document; specific assignments are tracked in Milo IPAM.
+
+The ULA pool (`fc00::/8`) is used exclusively for private tenant networking. These prefixes are never exposed to the underlay, never advertised to transit providers, and never appear in the global DFZ.
```
0xmc: If we do decide to go the ULA path, we should use `fd00::/8` as it may help avoid collisions in the cases of tenant-onboarding, M&A, or multi-cloud overlays.

slindseysr:
We assigned together a single PI /48 for loopbacks few days ago but I am thinking now we should assign the loopbacks out of the site specific unicast blocks per site.  This way the parent is unicast advertised for the site and thus loopbacks can intercommunicate in all scenarios (even over transit).  If we keep the single /48 and anycast it then the more specifics we require common backbone to intercommunicate.  Once we have our own backbone its less of an issue, but I am just thinking of scenarios where we are still straddling hosted infrastructure.

enhancements/networking/addressing_plan/README.md
```diff
@@ -0,0 +1,156 @@
...
+├── /48  ULA private VPC space       (from fc00::/8)
+├── /48  SRv6 SID locator space      (from [LOCATOR-BLOCK]/33)
+└── /48  Infrastructure loopbacks    (from [INFRA-BLOCK]/33)
+         ├── /49  Underlay BGP router loopbacks
+         └── /49  Overlay BGP router loopbacks
+```
+
+Per-PoP address assignments (actual prefix values) are tracked in Milo's IPAM service. This document defines the allocation scheme; Milo is the source of truth for specific assignments. See: [datum-cloud/enhancements#715](https://github.com/datum-cloud/enhancements/issues/715).
+
+### ULA /48 — Private VPC Addressing
```
0xmc: Do we want to use ULA as proposed or GUA? GUA does provide guarantees for uniqueness and expected routing family selection preferences that ULA does not. Regardless of ULA vs GUA, we should document why we're choosing one over the other.

enhancements/networking/addressing_plan/README.md
```diff
@@ -0,0 +1,156 @@
...
+
+Each PoP's ULA `/48` is the pool from which tenant VPC prefixes at that PoP are allocated. This space is entirely internal to the overlay and is never reachable from outside it.
+
+From the PoP's `/48`, each consumer VPC at that PoP is allocated a `/64`. A `/48` provides 65,536 possible `/64` allocations — this is the practical capacity ceiling for tenant VPCs at a single PoP, not a hard platform limit. The platform tracks allocations by association (VPC ↔ location) in the IPAM service, not by encoding topology in address bits.
```
0xmc: If the tenant has a VPC that spans multiple PoPs, their VPC will be comprised on non-aggregated /64s. This differs from the typical default one-CIDR-per-VPC which can then be subnetted per region/AZ/PoP. It is, however, possible I am missing a critical part of the design philosophy.

enhancements/networking/addressing_plan/README.md
privateip: Agreed, this should be 16 bits for the function block no 32.

enhancements/networking/addressing_plan/README.md
privateip: padding was assumed but will update the document to make it explicit

enhancements/networking/addressing_plan/README.md
privateip: i have no reservations about using the GUA address space in lieu of the ULA space if that is the consensus.   I can update the document to provide an either/or option.  given this is a design document and not an implementation document.

enhancements/networking/addressing_plan/README.md
privateip: i think using either address space is acceptable and will update the design document to provide the option to use either one

enhancements/networking/addressing_plan/README.md
privateip: each instance in gVPC is allocated a /128 out of the address space.  we aren't assigning full subnets by default.  we will have an option to allow a customer to design their own subnetting in gVPC as a feature in which case the customer can design a contiguous block if they want.

scotwells: A /128 would be default behavior, correct? We'll want an instance to support claiming larger subnet blocks so it can handle assigning out IPs to things on the host.

privateip: @scotwells correct

enhancements/networking/addressing-plan/README.md
```diff
@@ -0,0 +1,257 @@
...
+├── /48  ULA private VPC space       (from fd00::/8)
+├── /48  SRv6 SID locator space      (from [LOCATOR-BLOCK])
+└── /48  Infrastructure loopbacks    (from [INFRA-BLOCK])
+         ├── /49  Underlay BGP router loopbacks
+         └── /49  Overlay BGP router loopbacks
+```
+
+Per-PoP address assignments (actual prefix values) are tracked in Milo's IPAM service. This document defines the allocation scheme; Milo is the source of truth for specific assignments. See: [datum-cloud/enhancements#715](https://github.com/datum-cloud/enhancements/issues/715).
+
+### VPC Address Space Options
```
0xmc: In both the ULA and GUA options, do we want to support per-gVPC aggregation for tenants spanning more than one region/PoP?

0xmc:
@privateip looking good.  A few things that this plan doesn't address or are unclear:

1. I'm confused on VRF ID to prefix mappings.  VPCAttachment ID, IPAM, or L3VPN type-5?
2. From where are Compute, management, and other subnets numbered?  Are these just special purpose VPCs for Datum numbered out of the same space we use for customers?
3. From where are backbone links numbered?  The same block for loopbacks?

privateip:
@0xmc I have significantly updated the document based on your comments above.  I actually have split the document into two, one that focuses solely on IPv6 addressing and one that focuses on how SRv6 SIDs are defined and used.

Looking forward to your comments and/or approval.

enhancements/networking/srv6/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+---
+
+## Argument Semantics: VRF ID
+
+The 16-bit slot immediately following the active behavior uSID carries the VRF ID. VRF IDs are **PoP-local**; the same value may be reused across different PoPs because the 48-bit uSID block ensures global uniqueness.
+
+VRF ID `0x0000` is reserved and must not be allocated.
+
+The platform allocates VRF IDs sequentially within each PoP's namespace when a tenant VPC is first instantiated at that PoP. The platform tracks the mapping of tenant VPC to VRF ID independently per PoP.
```
0xmc: We may want to have a decommission/reuse policy defined.

enhancements/networking/ipv6/README.md
```diff
@@ -0,0 +1,265 @@
...
+> | ARIN     | North America                     |
+> | RIPE NCC | Europe, Middle East, Central Asia |
+> | APNIC    | Asia-Pacific                      |
+> | LACNIC   | Latin America and the Caribbean   |
+> | AFRINIC  | Africa                            |
+>
+> The specific RIR used depends on the platform's operational region and upstream carrier relationships.
+
+> [!NOTE]
+> As the platform scales, additional RIR PI blocks must be obtained for both the SRv6 locator and infrastructure loopback pools. Initiate RIR procurement well before existing allocations are exhausted; lead time for PI space is measured in weeks to months.
```
0xmc: Do you have a sense of initial assignments?  /32s?

0xmc (dismissed):
I left two comments, which should be resolved but aren't preventing me from approving.  We may also want to consider being opinionated about where in a block a sub-allocation comes from (top vs bottom) before getting into IPAM/control-plane design.  Again, not preventing me from approving.  Looks great.

privateip:
Flagging a citation issue surfaced while evaluating a downstream implementation plan based on this doc.

**RFC 9374 doesn't define this.** RFC 9374 is "DRIP Entity Tag (DET) for Unmanned Aircraft System Remote ID" (March 2023, DRIP WG) — a drone-identity spec, unrelated to SRv6/networking. Verified directly against rfc-editor.org/rfc/rfc9374 and datatracker.ietf.org/doc/rfc9374.

The relevant IETF documents are:
- **RFC 9800** — "Compressed SRv6 Segment List Encoding" (June 2025) — the actual multi-vendor consensus standard for SRv6 segment-list compression (what's colloquially called "uSID"/C-SID). Cisco's original standalone `draft-filsfils-spring-net-pgm-extension-srv6-usid` expired in June 2024 without reaching RFC status — it was never a ratified standard.
- **RFC 9252** — "BGP Overlay Services over SRv6" — the direct SRv6-native equivalent of RFC 4364 VPNs. Worth a look: it already defines the Service SID itself as the VRF/service identifier, locally significant at the egress node, with no separate decoded argument field required.

One more finding relevant to the VRF ID / Active uSID design in this doc: RFC 9800's compressed VRF-selection behaviors (`uDT`/REPLACE-CSID for End.DT4/DT6/DT46) explicitly state the Argument value is *ignored* by the endpoint — VRF selection happens via table lookup on the whole compressed SID, not by arithmetically decoding the argument bits. Separately, RFC 9800 scopes CSID reuse per Locator-Block (i.e. per site), specifically so a 16-bit space doesn't get exhausted platform-wide. Worth checking the Active-uSID registry section against that scoping, since a flat global registry shared across every PoP forgoes the per-site reuse the standard is designed around.

0xmc (dismissed):
I should have left my early comment as part of a review:

> I left two comments, which should be resolved but aren't preventing me from approving. We may also want to consider being opinionated about where in a block a sub-allocation comes from (top vs bottom) before getting into IPAM/control-plane design. Again, not preventing me from approving. Looks great.

enhancements/networking/ipv6/README.md
```diff
@@ -0,0 +1,283 @@
...
+
+#### GUA Tenant Addressing Policy
+
+These rules are mandatory whenever GUA VPC addressing (Option B) is in use. They apply to both platform-allocated GUA space and BYOP tenant space. Failure to enforce them will result in tenant prefixes entering the global DFZ.
+
+**Default deny export.** All prefixes that fall within the GUA VPC block are denied export at every underlay-facing BGP session by default. This is a BGP outbound prefix list applied at the session level, not a route-map condition that depends on community or attribute matching. For this control to be effective, the deny term must be evaluated after all allowlist permit terms in the session's export policy chain, and no other export policy applied to the same session may contain a terminating accept for a GUA VPC prefix based on other attributes — operators must verify per-platform that no earlier-evaluated policy term can produce a terminating permit before the deny is reached. There is no opt-out for individual sessions.
```
slindseysr: The planned export policy facing external peers meets the criteria stated here as only permits explicitly defined criteria and ends with explicit DENY everything term.  However, we will enhance with a DENY BOGONs term in leading term set.

enhancements/networking/ipv6/README.md
```diff
@@ -0,0 +1,283 @@
...
...refix is accepted into the overlay, the platform must verify: (1) the tenant's RIR allocation record (ARIN, RIPE, APNIC, or equivalent) confirms the prefix is theirs, (2) the tenant demonstrates current control of the prefix via a platform-issued challenge — publishing a platform-provided token or certificate in the prefix's RDAP record (analogous to AWS's X.509-certificate-in-RDAP plus signed authorization message), creating a PTR record under the prefix resolving to a platform-issued token (analogous to GCP's reverse-DNS validation), or a signed Letter of Authorization on company letterhead (as commonly required by transit providers and Cloudflare's BYOIP program), (3) the prefix is not currently being announced in the global DFZ by another origin AS (a migration/conflict-safety check, not itself an ownership proof), and (4) a valid ROA exists. The maximum prefix length for a BYOP allocation is the allocation's RIR-registered length — sub-allocations that are more specific than the registered block are not accepted without additional RIR evidence. BYOP validation is re-run on a scheduled basis; if a tenant's RIR record lapses, the BYOP prefix is quarantined pending re-validation.
```
slindseysr: Our route filtering arrangement with upstream providers will be purely based on our AS-SET and thus properly registered IRR objects are required to work on our platform.  We should not offer an LOA process as alternative.

enhancements/networking/ipv6/README.md
```diff
@@ -0,0 +1,283 @@
...
+
+**RPKI and ROA requirements.** The operator must publish ROAs for the platform's GUA VPC block with `maxLength = 48`. Per RFC 6483's ROA validation algorithm, this ensures any advertisement of a more-specific prefix (e.g., an individual `/64`) is classified RPKI Invalid; whether an Invalid route is actually dropped or merely deprioritized is a matter of each receiving peer's local policy (RFC 6811 §5 permits either), though in practice many transit and peering networks do filter Invalid routes outright, so this ROA still functions as a meaningful out-of-band backstop against more-specific leaks — treat it as defense-in-depth alongside the export prefix-list and route-leak-detection controls, not a substitute for them. For BYOP prefixes, the tenant must hold a valid ROA that either (a) origin-authorizes the platform's overlay ASN or (b) origin-authorizes the tenant's own ASN if the tenant manages route origination. Prefixes with RPKI Invalid status must not enter the overlay and must not be exported under any circumstances.
```
slindseysr: Agreed that we should explicitly drop routes invalid state routes.  However, we may want to consider allowing unsigned routes (state unknown) as RPKI adoption amongst operators is still progressing.

enhancements/networking/srv6/README.md
```diff
@@ -0,0 +1,253 @@
...
+
+---
+
+## Control Plane Mechanics (BGP EVPN L2VPN)
+
+Tenant routing and bridging states are distributed across the PoP fabric via BGP EVPN under the L2VPN EVPN address family (AFI=25, SAFI=70). The control plane maintains a deterministic mapping between BGP route identifiers and SRv6 VRF/EVI IDs, ensuring consistent SID construction at every ingress node.
+
+### Instance ID Signaling
+
+The VRF ID or EVI ID is signaled as part of the BGP EVPN Route Distinguisher (RD). Per RFC 4364 §4, an RD is an 8-byte value composed of a 2-byte Type field and a 6-byte Value field. The platform uses RD Type 1 (RFC 4364 §4.2), whose 6-byte Value field consists of a 4-byte Administrator subfield (originating node's own loopback IP) and a 2-byte Assigned Number subfield carrying the Instance ID (bounded to `1–4095` to match the C-SID Argument capacity):
```
0xmc: @privateip this means we're going to need IPv4 loopback addresses ("4-byte Administrator subfield (originating node's own loopback IP") unless we do something like https://blog.apnic.net/2026/06/12/bgp-router-id-structuring-in-ipv6-native-networks/

0xmc (commented):
I made a comment in diff for a specific part of the design.  Here are too less-specific comments.

### Instance ID discrepancies

Apologies for posting this without fully understanding the problem (not enough mental bandwidth today), but I think I reduced the slop to a manageable amount:

> - Line 50: "...a 12-bit Instance ID field (supporting up to 4,095 dynamic VRF/EVI IDs per node)..."
> - Line 130: "...4,095 usable identifier values per PoP (value 0x000 is reserved) — this bounds the number of distinct tenant routing/bridging instances addressable within a uSID Block..."
> - Line 191: "...a single /48 uSID block provides 4,095 usable VRF IDs... This limit applies per uSID Block, not per node or per PoP..."
> - Line 243: (in "Alternatives Considered") "...Constrains the Instance ID namespace to 12 bits (maximum 4,095 VRF/EVI IDs per node)."
> 
> A three-way inconsistency in phrasing: "per node" (L50, L243), "per PoP" (L130), and "per uSID Block, explicitly not per node or per PoP" (L191).
> 
> L191 is the most careful and is almost certainly the intended meaning — it's the one that accounts for the "Secondary uSID Block allocation" scaling path later in the same doc, where a single PoP can have more than one uSID Block. Under that scaling model, "per PoP" (L130) becomes wrong the moment a second block is added, and "per node" (L50/L243) was never right since all nodes in a PoP share the same Instance ID namespace per block.
> 

### PR description

This just needs an overall update to reflect the current state of the design.

enhancements/networking/srv6/README.md
privateip: Good catch, this should be a RD type 2 not type 1.  I will update the document

privateip:
One more pass at both documents.  I noticed there were operational concerns sneaking into the documents so I removed those and/or re-worded them to be more design oriented.   This latest revision should address all outstanding concerns.

cc @slindseysr @0xmc

0xmc (dismissed):
Looks great to me.  If any (literal) bits are slightly off, they'll be caught in implementation.  I see no reason not to move forward at the plan at this juncture.

slindseysr (approved):

PR #742, opened 2026-05-25
feat: distributed inference edge

drewr:
## Summary

Proposes a distributed LLM inference network spanning Datum's global edge PoP footprint. Adapts the architecture from https://gist.github.com/drewr/7a061d09377de996a3f4129eaed5c635 to Datum's specific infrastructure.

## What this does

The enhancement describes serving OpenAI-compatible LLM inference across Datum's 20+ edge PoPs by treating inference as a distributed systems problem rather than an ML problem. It reuses infrastructure Datum already operates:

- **SRv6 / Galactic VPC** for cross-PoP session steering (roaming clients transparently follow their KV cache state without HTTP redirects)
- **Envoy** (already at edge clusters) for within-PoP GPU-aware load balancing via xDS
- **DCGM / Prometheus** for VRAM pressure metrics driving load balancing weights
- **Datum Workloads API** for inference node scheduling with GPU accelerator resources and VRAM-based autoscaling
- **FluxCD** GitOps for deployment following the existing edge cluster pattern

## New components introduced

- `pop-router`: xDS control plane per PoP, drives Envoy cluster config from DCGM metrics
- Redis per PoP: local session registry and KV cache index
- Global control plane (Phase 3): anycast weight control, prefix cache promotion, cross-PoP SRv6 policy provisioning

## Build phases

1. Single PoP, GPU-aware Envoy load balancing — fully functional and observable
2. Second PoP + SRv6 session steering — transparent client roaming
3. Automated global control plane — anycast weights, prefix cache optimization
4. (Optional) Pipeline parallelism for oversized models

## Dependencies

- Galactic VPC underlay network requirements (MTU ≥ 9080 on inter-PoP paths, IPv6 full mesh) must be satisfied before Phase 2
- GPU-capable edge cluster node pools (GKE) provisioned via infra-provider-gcp

## Location

`enhancements/inference/distributed-inference-edge/README.md`


drewr:
Review findings + C4 diagrams — commit [`6f87db6`](https://github.com/datum-cloud/enhancements/commit/6f87db6)

Reviewed the proposal for architecture gaps, security issues, and things that won't work as described. Addressed everything in the commit above. Summary of what changed and why:

---

### 🔴 Fixes for things that wouldn't work

**Session ID transport was unspecified.** The roaming story depends on PoP B knowing the client's session ID on reconnect, but the OpenAI API has no session ID field and the doc was silent on the mechanism. Defined `X-Datum-Session-ID` as the carrier: issued by `pop-router` on the first response, echoed by the client on reconnect. Updated the session routing description and the user story to match. The dynamic C4 diagram shows the 12-step roaming flow with this spelled out.

**`cgroup limits prevent VRAM-exhausted pods from impacting other workloads` is incorrect.** GPU VRAM is not controlled by Linux cgroups — `cudaMalloc` bypasses cgroup memory accounting entirely. Corrected the risk mitigation to say what we actually rely on: dedicated GPU node pools (no shared capacity) + DCGM-driven health checks that pull a node from Envoy rotation before exhaustion. Added a note that NVIDIA MIG is available for Phase 2 if multi-tenant shared hardware is ever needed.

**Redis per PoP was a single point of failure.** A single Redis instance going down takes out all session affinity and KV cache routing at that PoP, and the "graceful degradation" framing understates the thundering-herd risk on recovery. Upgraded to a 3-node Redis Sentinel cluster per PoP.

---

### 🟡 Design gaps addressed

**Model cold-start makes the autoscaling strategy risky.** A 70B model at 4-bit takes 2–5 minutes to load into VRAM. Without a readiness gate, `pop-router` could add a scaling node to Envoy before the model is loaded. Added a model readiness gate (init container / sidecar) that blocks the readiness probe until VRAM warmup completes. Also noted that autoscaling thresholds need to trigger *before* headroom is critically low, not in reaction to it.

**Phase 2 session registry topology was undefined.** "Lightweight" with no topology means unknown reliability characteristics and an undefined migration path to the Phase 3 consensus store. Specified single-region CockroachDB (3-node quorum) for Phase 2, with Phase 3 expanding to multi-region — no data migration required.

**`InferenceSession` at millions of objects doesn't belong in etcd.** etcd is not designed for millions of short-lived application-layer records at high write rates. The `InferenceSession` Kubernetes API surface will be backed by a CockroachDB aggregated API server, following the existing Datum pattern.

**Cross-PoP cache transfer threshold used "expected generation length" — which is unknowable at request time.** Replaced with a concrete heuristic using `max_tokens` from the request as a proxy.

**`:latest` image tags in the Workload spec example.** Pinned to semver/digest. Small thing but it sets the pattern for what operators copy-paste.

---

### 🟠 Security additions

Added a Security section covering three things that were absent:

- **Auth**: inference API sits behind the existing Datum API gateway for key/JWT validation and rate limiting; inference nodes are not publicly reachable.
- **Cross-PoP KV cache transport**: KV cache shards contain embeddings derived from user prompts. Added mTLS at the application layer for `inference.cache-xfer` transfers (fabric encryption alone is not sufficient). All `pop-router` ↔ global control plane and Redis Sentinel cluster traffic also uses mTLS.
- **Tenant isolation**: KV cache keys are now scoped as `tenant_id:SHA-256(prefix)`. Added a note on why cross-tenant timing-based cache oracle attacks are mitigated.

---

### 📐 C4 diagrams

Four diagrams added alongside the README, rendered with the Datum PlantUML theme:

- **System Context** — the inference network in its environment
- **Network Topology** — the CDN-analog view: anycast entry → edge PoPs → SRv6 backbone → global control plane
- **PoP Components** — every component inside a single PoP with data-flow relationships
- **Session Roaming Flow** — the 12-step cross-PoP reconnect sequence, with annotations on the session ID gap and the ~1 WAN RTT latency budget

PR #745, opened 2026-05-28
Proposed POP design enhancements

slindseysr:
The README here outlines current gaps in current edge deployment and proposes improvements to include high level detail for colo deployment.


enhancements/networking/pop-enhancement/README.md
```diff
@@ -0,0 +1,135 @@
...
+\
+![](./image2.png)
+
+**Target Architecture 2 (TA2): Dedicated Routers**
+
+Like TA1, TA2 also solves all of the identified gaps from the current
+architecture. The difference is that it also addresses the identified
+gaps/weakness from TA1. The key change is moving the external edge
+routing and forwarding functions from FRR/VPP to dedicated routers
+(preferably Juniper). The routers will be dual homed to the TORs in the
```
0xmc: I'm curious why the TORs are on the north side of the routers.

slindseysr: Thanks for the review and feedback.  Logically they are at the top but yeah physically not so much.  I should describe better the why in the proposal.  Some motivating factors below.

- Easy drop in retrofit TA1 to TA2.  No need touch any facility facing patches.  Just connect the routers and move the trunking
- Conserve limited ports on the routers. We have way more ports on the switches than we will ever need.
- Flexibility to connect any circuit to either router.  We will try to connect 1 of everything to each TOR but there will be cases where we only order 1 port due to cost.  Good example is expensive IX.  Connect it to TOR1 but we can trunk it to either router we choose.  If a router fails or we want to do maintenance we can logically move it to the other router.

privateip:
Fundamentally I have no issues or concerns with the proposed designs.   I am a little concerned that this is a bit disjointed from the gVPC designs as documented [here](https://github.com/datum-cloud/enhancements/pull/745) and [here](https://github.com/datum-cloud/enhancements/pull/705).   I think we need to take some time to bring the two efforts together.  If we dont, we will end up with two purpose built networks, one for gVCP and one for Anycast.

privateip (changes requested):
It would be helpful if this document listed out the goals and non-goals of the desired final state.  It kind of touches on this in but a dedicated list of achievable goals and defined non-goals would be helpful imho

slindseysr:
We met and aligned on go the forward designs.  Closing this PR to defer to gVPC references shared in thread

Issue #698, opened 2026-04-23
Headless experience for Datum tunnels


privateip:
### High-Level Summary

Add a headless CLI that creates and manages tunnels to Datum Cloud without requiring a graphical interface. The CLI can run as a foreground process or be daemonized as a background process, and optionally install a tunnel as a system service (systemd on Linux, launchd on macOS, Windows Service on Windows) for persistent, boot-time tunnel management — mirroring the service lifecycle model familiar from Docker and Podman.

### Motivation

The Datum desktop app currently requires a display environment, which excludes a large class of legitimate use cases:
* headless servers,
* CI/CD pipelines,
* container workloads,
* SSH-only remote machines, and
* automated infrastructure

where no GUI is available or desirable. Operators who want a persistent tunnel on a server today have no supported path. It's either run the GUI app in a degraded state, cobble together workarounds, or don't use Datum at all. The service installation feature directly addresses the operational gap where tunnels need to survive reboots and run unattended, the same problem Docker and Podman solve for container workloads with systemctl.

Without a CLI and service model, Datum is positioned as a developer convenience tool rather than production infrastructure.

### Goals

- Users can invoke datum tunnel (or equivalent subcommand) from any terminal to establish a tunnel without a display environment
- The CLI accepts all tunnel configuration options currently available in the desktop UI (auth, endpoint, credentials, network settings, etc.)
- Users can run the tunnel in the foreground (blocking, logs to stdout/stderr) or detach it as a background daemon
- Users can install a named tunnel as a persistent system service (datum tunnel install --name <name>) that auto-starts on boot, using systemd on Linux, launchd on macOS, and Windows Service Manager on Windows
- Service lifecycle management commands are supported: install, uninstall, start, stop, status
- The CLI exits with standard POSIX exit codes so it composes cleanly with shell scripts and process supervisors

### Non-Goals

- This does not replace or deprecate the existing desktop GUI, the GUI remains the primary UX for interactive desktop users
- No new tunnel types or networking capabilities are introduced; the CLI exposes existing functionality only
- Container-native packaging (e.g., an official Docker image with an entrypoint) is out of scope for this enhancement, though the headless CLI makes it straightforward to build one separately
- Cross-platform service installation is not required to ship simultaneously, Linux/systemd ships first; macOS and Windows follow
- Remote management of tunnels (controlling a daemon over a network socket or API) is out of scope

richardhenwood:
[...]

Issue #130, opened 2025-06-16
Create auth-provider-zitadel Repository and Bootstrap Project


JoseSzycho:
### High-Level Summary

Establish a new repository, `auth-provider-zitadel`, to house the integration logic, controllers, and API client for Zitadel as the authentication provider for machine accounts and machine account key management.

### Motivation

A dedicated repository is needed to encapsulate all Zitadel-specific logic and controllers, making it easier to maintain, test, and evolve the integration independently from the main milo codebase.

### Goals

- Create the `auth-provider-zitadel` repository.
- Bootstrap the project with initial structure, dependencies, and CI setup.
- Document the repository’s purpose and usage.

### Non-Goals

- Implement any business logic for machine account or key management.
- Integrate with the `mil` repo at this stage.

JoseSzycho:
`auth-provider-zitadel` repo: https://github.com/datum-cloud/auth-provider-zitadel

PR #756, opened 2026-06-05
feat: add Headless Tunnel CLI enhancement

drewr:
## Summary

Adds a provisional enhancement document under `enhancements/networking/connectors/headless-tunnel-cli/` for a headless command-line experience for Datum tunnels.

The enhancement covers:
- A new `datumctl connect tunnel ...` plugin that wraps the existing `datum-connect` Rust binary.
- Three execution modes against the same tunnel runtime:
  - **Foreground** (blocking, logs to stdout/stderr)
  - **Background daemon** (detached process, PID + log files, managed with `tunnel ps`/`stop`/`logs`)
  - **System service** (systemd / launchd / Windows SCM, via `kardianos/service`)
- Service-account auth integration through datumctl's existing credentials helper, so long-running unattended tunnels refresh credentials automatically.
- POSIX-conformant exit codes, named tunnels, user-vs-system scope, and a full PRR questionnaire.
- Phased platform delivery: Linux/systemd in v0.1, macOS/launchd in v0.2, Windows SCM in v0.3.

Refs datum-cloud/enhancements#698

## Status

`provisional` / `alpha` / `latest-milestone: v0.1`. Submitting per the template guidance to "merge early and iterate" — feedback in subsequent PRs welcome.

## Test plan

- [ ] Reviewers confirm the goals/non-goals match the intent in datum-cloud/enhancements#698 and the discussion thread.
- [ ] Reviewers validate the auth model (datumctl helper + service accounts) is the right unification point versus alternatives listed in §Alternatives.
- [ ] Reviewers confirm the location under `networking/connectors/` is appropriate (sibling to `initial-proposal/`).
- [ ] Reviewers flag any PRR questions that need stronger answers before moving to `implementable`.

### DIFF
--- a/enhancements/networking/connectors/headless-tunnel-cli/README.md
+++ b/enhancements/networking/connectors/headless-tunnel-cli/README.md
@@ -333,6 +333,21 @@ This is what `HeartbeatAgent` and the kube-client portions of the runtime
 exist to do; they are not optional plumbing — without them, the gateway has
 nothing to dial.
 
+The TTL on the TXT record (`IrohConnectorConfig.TTLSeconds` in NSO,
+default 30) is load-bearing for restart UX: if a tunnel restarts before
+the previous record's TTL expires, the gateway's resolver returns the
+stale answer and the new endpoint isn't reachable until it expires.
+Measured behaviour:
+
+| Scenario | Time from `datum-connect` start to proxy responding |
+|---|---|
+| Restart immediately after stop | ~30s (TTL-bound) |
+| Restart 35s after stop (TTL elapsed) | ~1s |
+
+The 30s floor is essentially the cached TTL counting down; control-plane
+plumbing (PCP heartbeat → NSO reconcile → DNSRecordSet → Envoy
+metadata) completes in well under a second in both cases.
+
 ### Execution Modes
 
 ```
--- a/enhancements/networking/connectors/headless-tunnel-cli/README.md
+++ b/enhancements/networking/connectors/headless-tunnel-cli/README.md
@@ -1,0 +1,923 @@
+---
+status: provisional
+stage: alpha
+latest-milestone: "v0.1"
+---
+
+# Headless Tunnel CLI
+
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+  - [Notes/Constraints/Caveats](#notesconstraintscaveats)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [Execution Modes](#execution-modes)
+  - [Architecture: Go Plugin + Rust Binary](#architecture-go-plugin--rust-binary)
+  - [Auth Integration](#auth-integration)
+  - [Long-Running Token Refresh](#long-running-token-refresh)
+  - [Command Surface](#command-surface)
+  - [Background Daemon Mode](#background-daemon-mode)
+  - [System Service Installation](#system-service-installation)
+  - [State Isolation](#state-isolation)
+  - [Exit Codes](#exit-codes)
+  - [Required Library Changes](#required-library-changes)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+  - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
+  - [Monitoring Requirements](#monitoring-requirements)
+  - [Dependencies](#dependencies)
+  - [Scalability](#scalability)
+  - [Troubleshooting](#troubleshooting)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Infrastructure Needed](#infrastructure-needed)
+
+## Summary
+
+Today, the only supported way to run a Datum tunnel is through the Datum
+desktop application, which requires a display environment. This enhancement
+adds a headless command-line experience for tunnels, distributed as a
+[datumctl][datumctl] plugin (`datumctl connect tunnel ...`). The plugin
+supports three execution modes against the same underlying tunnel runtime:
+
+1. **Foreground** — blocks until interrupted, logs to stdout/stderr.
+2. **Background daemon** — detached process, PID file, log file, managed with
+   `tunnel ps` / `tunnel stop` / `tunnel logs`.
+3. **System service** — registered with systemd (Linux), launchd (macOS), or
+   Windows Service Manager so tunnels survive reboot and run unattended.
+
+For unattended modes, authentication is provided by Datum Cloud service
+accounts (`datumctl login --credentials key.json --session <name>`), which
+datumctl already supports. The plugin invokes datumctl's credentials helper
+for both initial token acquisition and long-running token refresh, so a single
+installed tunnel can run for days or years without manual intervention.
+
+The plugin wraps the existing `datum-connect` Rust binary, which holds the
+iroh data plane and the K8s control-plane integration. The Go plugin layer
+handles process lifecycle, configuration persistence, and service-manager
+integration; the Rust binary handles the tunnel itself.
+
+[datumctl]: https://github.com/datum-cloud/datumctl
+
+## Motivation
+
+The Datum desktop application requires a display environment. That excludes
+a large set of legitimate Datum use cases:
+
+- Headless servers (cloud VMs, on-prem infrastructure)
+- CI/CD pipelines
+- Container workloads
+- SSH-only remote machines
+- Automated infrastructure
+
+Operators who want a persistent tunnel on a server today have no supported
+path. They either run the GUI app in a degraded mode, cobble together
+workarounds, or do not use Datum at all. The system-service capability
+directly addresses the operational gap where tunnels need to survive reboots
+and run unattended — the same problem Docker and Podman solve for container
+workloads with `systemctl`.
+
+Without a headless CLI and service model, Datum is positioned as a developer
+convenience tool rather than as production infrastructure.
+
+### Goals
+
+- Users can invoke `datumctl connect tunnel ...` from any terminal to manage
+  tunnels with no display environment.
+- The CLI exposes every tunnel configuration option the desktop UI exposes
+  today (label, endpoint, enabled state).
+- Users can run a tunnel in the foreground (blocking, logs to stdout/stderr)
+  or detach it as a background daemon.
+- Users can install a named tunnel as a persistent system service that
+  auto-starts at boot, using the OS-native service manager (systemd on Linux,
+  launchd on macOS, Windows Service Manager on Windows).
+- Service lifecycle management commands are supported: `install`,
+  `uninstall`, `start`, `stop`, `status`.
+- The CLI exits with standard POSIX exit codes so it composes cleanly with
+  shell scripts and process supervisors.
+- Long-running tunnels refresh credentials automatically and indefinitely.
+- A single named tunnel can run for days, weeks, or years without manual
+  re-auth.
+
+### Non-Goals
+
+- This does not replace or deprecate the existing desktop GUI. The GUI
+  remains the primary UX for interactive desktop users.
+- No new tunnel types or networking capabilities are introduced. The CLI
+  exposes existing functionality only.
+- Container-native packaging (an official OCI image with an entrypoint) is
+  out of scope for this enhancement, though the headless CLI makes it
+  straightforward to build one separately. See
+  [Implementation History](#implementation-history) for a follow-up note on
+  container distribution.
+- Cross-platform service installation is not required to ship simultaneously.
+  Linux/systemd ships first; macOS/launchd and Windows SCM follow in
+  subsequent milestones.
+- Remote management of tunnels (controlling a daemon over a network socket
+  or API) is out of scope.
+- Identity material baked directly into a per-user binary download
+  ("Identity Packet") is out of scope for the MVP; see
+  [Alternatives](#alternatives).
+
+## Proposal
+
+A new datumctl plugin, `datumctl-connect`, ships as a release of this
+repository and is registered in
+[datum-cloud/datumctl-plugins](https://github.com/datum-cloud/datumctl-plugins).
+The plugin binary bundles the existing `datum-connect` Rust binary in the
+same release tarball. The Go plugin handles command-line surface, process
+lifecycle, and service-manager integration; the Rust binary runs the actual
+tunnel.
+
+The Rust binary is refactored to support a "plugin mode" in which it
+receives its bearer token via env var and renews it by invoking datumctl's
+credentials helper — never opening a browser, never running OIDC discovery,
+never writing OAuth state to disk. The datumctl credentials helper supports
+both interactive sessions and Datum Cloud service-account sessions
+transparently, so the same code path serves long-running unattended tunnels.
+
+### User Stories
+
+#### Story 1: Developer runs a quick tunnel from a terminal
+
+> As a developer, I want to expose my local dev server to the Datum network
+> from my terminal, without launching a desktop app.
+
+```sh
+datumctl connect tunnel listen --label my-dev --endpoint 127.0.0.1:3000
+```
+
+The tunnel blocks the terminal, prints the assigned hostname, and tears down
+cleanly on Ctrl+C.
+
+#### Story 2: Operator runs a background tunnel on a workstation
+
+> As an operator, I want to start a tunnel that survives me closing my
+> terminal, but doesn't require root or a service installation.
+
+```sh
+datumctl connect tunnel listen --detach --name dev-app \
+  --label dev-app --endpoint 127.0.0.1:3000
+datumctl connect tunnel ps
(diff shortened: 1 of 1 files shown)


drewr:
## Design review — ambiguities and contradictions

Strong overall direction, especially the "auth is owned by datumctl" constraint. A few things to clarify before this moves from `provisional` to `implementable`. Grouped by severity.

### Contradictions (need a clear decision)

**1. `listen_key` storage layout conflicts with the just-landed per-project layout.**

§State Isolation says:
> Each named tunnel has its own subdirectory (`<repo>/tunnels/<name>/`) holding its own `listen_key` and proxy state.

But the headless tunnel work in [datum-cloud/app@76987b7](https://github.com/datum-cloud/app/commit/76987b7) moved `listen_key` to `<repo>/projects/<project_id>/listen_key` precisely to prevent cross-project iroh-pubkey collisions ([the diagnosis behind this lives in PR datum-cloud/enhancements#130](https://github.com/datum-cloud/app/pull/130#issuecomment-4643636460)). Two open questions:

- Should named tunnels nest *inside* projects (`<repo>/projects/<project>/tunnels/<name>/listen_key`) so the cross-project guarantee survives?
- Or does `--name` *imply* a project binding strong enough to replace the per-project scoping?

Either is defensible but the design has to pick one. As written, the plugin would regress the cross-project collision fix.

**2. The detach-mode "ready" definition contradicts the verify-phase work and the SLO.**

§Background Daemon Mode:
> The parent Go process waits up to 10 seconds for the tunnel to reach ready, prints the assigned hostname, and exits 0. If readiness takes longer, the parent exits 0 with a hint to run `tunnel status`.

But:

- §SLOs says: *"99% of `tunnel listen` invocations reach ready state in under 30 seconds"* — already longer than 10s.
- The verify phase landing in [datum-cloud/app@1ed969e](https://github.com/datum-cloud/app/commit/1ed969e) has a 10-minute default for a reason: data-plane reachability on resume routinely takes 1–3 minutes ([root cause traced in datum-cloud/iroh-gateway#12](https://github.com/datum-cloud/iroh-gateway/issues/12)).
- A parent that prints a hostname and exits 0 before the URL serves 2xx hands the user a URL that's broken-on-first-use ≥ 50% of the time today.

Pick one of: (a) wait for the full verify phase before exit; (b) exit *non-zero* (e.g. 75 — see contradiction #3) when readiness isn't confirmed within the window; (c) don't print the hostname yet, and require `tunnel status` to surface it once verified.

**3. Exit code for slow-ready detach contradicts the exit-code table.**

§Exit Codes defines:
> 75 — Temporary failure: tunnel exists but is still provisioning

But §Background Daemon Mode explicitly says the parent exits **0** when readiness exceeds the 10s window. Exiting 0 with "well, run `tunnel status`" is exactly the case 75 is designed for; the current design contradicts its own spec.

### Ambiguities (need spelling out)

**4. `--name` vs `--label` vs `--id` semantics aren't sharp.**

- User Story 2: `listen --detach --name dev-app --label dev-app` — same value used for both. Are they synonyms? If not, why is conflating them the canonical example?
- §Command Surface uses `--name N` for daemon process management and `--id X` for `update`/`delete`. We've already shipped `--id` in the standalone CLI as the *server-side* HTTPProxy resource name. The plugin's `--id` and `--name` must therefore be different things — local vs server-side identifiers. Spell this out.
- The picker UX shipped in [datum-cloud/app@7de50c7](https://github.com/datum-cloud/app/commit/7de50c7) and the auto-adoption in [@cff37e7](https://github.com/datum-cloud/app/commit/cff37e7) aren't mentioned anywhere. Do they survive in plugin mode? If yes, how do they interact with `--name`?

**5. `--project` is transient-override in some places and persisted-config in others.**

§Required Library Changes item 5:
> treats `--project` as a transient override (no `set_selected_context`)

But §System Service Installation persists `project: proj-prod` in the YAML at install time. So:

- At `install` time, `--project` is persistent.
- At `listen` time (foreground or detach), `--project` is transient.
- What if a user runs `tunnel start --name foo --project different-one` against an installed service? Does the override take effect, get ignored, or error out?

**6. The `DATUM_ACCESS_TOKEN` env var has unclear role.**

§Auth Integration:
> The Rust binary in plugin mode receives a fresh bearer token from the Go plugin via env (`DATUM_ACCESS_TOKEN`) at startup and renews it by execing `$DATUM_CREDENTIALS_HELPER`.

If the Rust process can exec the helper itself, what does `DATUM_ACCESS_TOKEN` achieve? Optimization (avoid one helper exec at startup)? Or is the Rust process *forbidden* from execing the helper until the initial token expires? The design also says the Rust process runs the refresh loop — so the Go plugin's role after exec is just "start the Rust binary." The env var feels redundant.

**7. `tunnel run` is "internal" but completely unspecified.**

§Command Surface lists `tunnel run --name N      # internal; invoked by service unit`. What does it accept? Does `--system` apply (the service unit args in §System Service Installation show it appears as `--system` for system-scoped services, but `run` itself isn't documented)? Where does it load config from when launched by systemd vs by `start`?

**8. Heartbeat scope in plugin mode is unspecified.**

We just shipped `HeartbeatAgent::start_manual()` ([datum-cloud/app@b7e9d6b](https://github.com/datum-cloud/app/commit/b7e9d6b)) so the CLI doesn't auto-enroll heartbeats for every accessible project. §Required Library Changes mentions `HeartbeatAgent::with_pcp_client(...)` but doesn't say whether plugin mode uses manual or auto-enroll. A named tunnel almost certainly wants manual mode (one project, no fan-out) — should be stated.

**9. "Control-plane dry-run" for install validation is hand-wavy.**

§System Service Installation install validations item 4:
> The service account has permission to create tunnels in the target project (control-plane dry-run).

A `SelfSubjectAccessReview` against the relevant verbs is the natural fit — but say so. A literal apiserver `dryRun=All` Create isn't quite the same thing (it would still consume a quota check, and quota check timeouts are flaky enough today to false-fail install).

**10. Two paths captured at install time, only one is documented.**

§System Service Installation says the absolute path to `datumctl-connect` is captured for the service unit. But the credentials helper is `datumctl` (a different binary). The persisted YAML (`<config-dir>/services/<name>.yaml`) doesn't show the helper path captured anywhere — yet §Risks claims:
> the plugin captures [the helper's] absolute path at install time

Either the YAML needs a `credentials_helper_path` field, or the design needs to clarify how the helper is found at runtime.

**11. `tunnel update` in plugin mode vs persisted service YAML.**

If I `install --name prod --endpoint 127.0.0.1:8080` and later `tunnel update --id prod-tunnel-xyz --endpoint 127.0.0.1:9000`, the server-side HTTPProxy backend changes but the persisted YAML at `<config-dir>/services/prod.yaml` still says `8080`. On next service restart, which one wins? Either the YAML must be re-read on each `run` (server is the truth) or the YAML must be updated by `update` (config is the truth). Pick one.

### Minor inconsistencies

**12. Container/sidecar scope statement is fuzzy.** §Non-Goals says OCI packaging is out of scope; §User Stories Story 5 shows the use case as "Future" and the design notes "the architecture supports it." This is fine narratively but a single sentence reconciling "out of scope for this PR" vs "designed to enable" would prevent reviewer confusion later.

**13. Windows/arm64 inclusion is contradictory.** §Risks treats it as contingent on CI passing. §Infrastructure Needed lists it as guaranteed. Pick one; if contingent, drop it from the matrix until the build is green.

**14. ApiEnv env-var conditional.** §Required Library Changes item 4: "honor `DATUM_API_HOST` in plugin mode." How does `ApiEnv` know which mode? Plugin detection lives in `cli/src/main.rs` (item 5) — but `ApiEnv` is in `lib/`. Without thread-through, this becomes "honor `DATUM_API_HOST` always" or requires a new constructor variant.

### Suggested follow-ups (not blocking)

- §Implementation History could pin the per-project-key migration ([@76987b7](https://github.com/datum-cloud/app/commit/76987b7)) and the idempotent-update_project ([@0311960](https://github.com/datum-cloud/app/commit/0311960)) as prerequisites — both are properties the plugin will rely on for "no resume churn" semantics, and §Required Library Changes doesn't mention them at all.
- A short §Compatibility section noting that the desktop app, standalone `datum-connect`, and the plugin will each register *distinct* Connector objects against the same iroh public key — unless they share a `Repo`. The cross-agent semantics deserve explicit treatment given how much pain the cross-project collision case caused this past week.

Otherwise: ready to start implementing the lib-side changes (#1, datum-cloud/cloud-portal#1335, #3 in §Required Library Changes are well-scoped and additive). I'll wait to do that until the contradictions above have a direction.

drewr:
## Resolution proposal — applying recent app-side precedents

Context from offline discussion: this enhancement was drafted before a week of intensive debugging in [datum-cloud/app#130](https://github.com/datum-cloud/app/pull/130) that surfaced several non-obvious requirements. The agreement is that **where today's app-side work conflicts with this enhancement, today's precedent wins** — the enhancement should evolve to incorporate it. The trajectory is: today's `app/cli/` will be split into (a) a binary in a new `connect` repo invoked by datumctl, and (b) a shared Rust crate in the same `connect` repo, used by the desktop app too.

Recording the resolution table and the lib-vs-binary boundary here so the design can absorb it.

### Resolutions for the ambiguities and contradictions flagged above

| # | Issue | Today's precedent | Resolution |
|---|---|---|---|
| 1 | `listen_key` layout (§State Isolation says `<repo>/tunnels/<name>/listen_key`) | [datum-cloud/app@76987b7](https://github.com/datum-cloud/app/commit/76987b7) — per-project keys at `<repo>/projects/<id>/listen_key`, motivated by cross-project iroh-pubkey collisions | Named tunnels become per-name repos: `<config-dir>/tunnels/<name>/` *is itself* a `Repo`. The per-project layout applies inside each named tunnel's repo. The plugin sets `DATUM_CONNECT_REPO=<config-dir>/tunnels/<name>` and `Repo` doesn't need to know about names. |
| 2 | Detach-mode "wait up to 10s for ready, then print hostname and exit 0" | [datum-cloud/app@1ed969e](https://github.com/datum-cloud/app/commit/1ed969e) — verify phase with default 10m, plus [datum-cloud/iroh-gateway#12](https://github.com/datum-cloud/iroh-gateway/issues/12) confirming routine 1–3 min iroh dial latency on resume | Drop the wait-then-print contract. Detach writes the PID immediately and exits with a "setting up in background; `tunnel status` will show progress" message. Verification continues in the daemon's log. No misleading early-success print. |
| 3 | Exit code 75 ("still provisioning") defined but never used | Same as datum-cloud/cloud-portal#1335 | Becomes correctly unused (or removed from the table). Slow-ready isn't a state the parent ever observes once detach exits immediately. |
| 4 | `--name` vs `--label` vs `--id` semantics | [datum-cloud/app@ca4470f](https://github.com/datum-cloud/app/commit/ca4470f) (`--id` = HTTPProxy resource name) + [@a68d8ae](https://github.com/datum-cloud/app/commit/a68d8ae) (`--id` + `--endpoint` must agree exactly) | `--name` is local-only (process management — PID file key, repo subdir name). `--id` is server-side. `--label` is decorative (display annotation). Document them as orthogonal. The picker + auto-adopt-single from [@7de50c7](https://github.com/datum-cloud/app/commit/7de50c7) / [@cff37e7](https://github.com/datum-cloud/app/commit/cff37e7) carry into the plugin's foreground `listen`. |
| 5 | `--project` "transient override" vs persisted in service YAML | Today's CLI requires a selected context at Tunnel handler entry | Persist `project` in the service YAML for installed services. Reject runtime override at `tunnel run`/`tunnel start` when it disagrees with the persisted value (fail loud, don't silently retarget). Foreground/detach: transient override is fine. |
| 6 | `DATUM_ACCESS_TOKEN` env var's role | n/a (new) | Drop it. The Rust process execs the helper at startup AND for refresh. One mechanism, less to keep in sync. Detect plugin mode via `DATUM_SESSION` (or an explicit `DATUM_PLUGIN_MODE=1`) instead of the presence of an initial token. |
| 7 | `tunnel run` subcommand surface | n/a | Spec it: `tunnel run --name N [--system]`. Loads the persisted YAML from the service dir. Refuses any other flags. The service unit literally writes that argv. |
| 8 | Heartbeat scope in plugin mode | [datum-cloud/app@b7e9d6b](https://github.com/datum-cloud/app/commit/b7e9d6b) — `start_manual()` for single-project | Plugin mode uses manual heartbeat scoped to the named tunnel's project. Drop the auto-enroll path for plugin entirely. Desktop app keeps `start()` because the multi-project UI actually wants that. |
| 9 | "Control-plane dry-run" install validation | n/a | Use `SelfSubjectAccessReview` against the required verbs (create on `connectoradvertisements`, `httpproxies`, etc.) Not `dryRun=All` Create — that would burn a quota-check round trip unnecessarily and is flaky enough today ([@d8f7c96](https://github.com/datum-cloud/app/commit/d8f7c96)) to false-fail install. |
| 10 | Helper path capture | n/a | Add `credentials_helper_path` to the persisted YAML alongside `name`/`project`/`session`. The service unit reads it from there at `run` time. Spell out that the plugin captures it once, at install. |
| 11 | `update` vs persisted YAML truth | [datum-cloud/app@0311960](https://github.com/datum-cloud/app/commit/0311960) — `update_project` is idempotent at the lib boundary; server-side is the authoritative state | Server is the source of truth for `label`/`endpoint`. YAML stores only what the *daemon* needs to start: `name`, `project`, `session`, `credentials_helper_path` (and the install metadata). `tunnel update` mutates the server only; `run` queries the server on startup for current label/endpoint. |
| 12 | Container/sidecar scope | n/a | Single sentence reconciling "out of scope for this PR; designed to enable." Resolve narratively, no behavior change. |
| 13 | windows/arm64 contingent in §Risks vs guaranteed in §Infrastructure | n/a | Drop from the v0.1 matrix until the build is green; add when verified. |
| 14 | `ApiEnv` honoring `DATUM_API_HOST` "in plugin mode" | n/a | Honor `DATUM_API_HOST` always (it's just an env var). Plugin-mode detection lives in the CLI layer, not in `ApiEnv`. |

### Lib-vs-binary boundary (the trajectory the rest of this work depends on)

Mapping today's `app/lib/` + `app/cli/` to the future `connect` repo:

**Shared crate `connect-lib` (Rust)** — consumed by the desktop app *and* by the plugin's Rust binary:

- `Repo` (per-project / per-named-tunnel scoping)
- `ListenNode`, `ConnectNode`, `Node`
- `TunnelService` (already idempotent at the lib boundary per [@0311960](https://github.com/datum-cloud/app/commit/0311960))
- `HeartbeatAgent` with both `start()` and `start_manual()`
- `ProjectControlPlaneClient`
- `TunnelProgress` + `ProgressStep` + `StepStatus`
- `TunnelService::get_active_progress` (the progress query)
- `await_tunnel_progress` (moved out of `cli/main.rs` — pure orchestration on top of `get_active_progress` with a callback-based progress reporter; no `println!` baked in)
- `verify_endpoints` (same — pure HTTP probe loop returning a `Result`; callers provide progress callbacks)
- `format_terminal_failure` (pure string formatter; fits in lib)
- New: `BearerSource` trait with two implementations:
  - `DatumCloudClient` (existing OAuth path — used by desktop)
  - `ExternalTokenSource` (helper-execed — used by plugin)

  `ProjectControlPlaneClient::new_with_bearer_source` gives both consumers a clean entry point.

**Binary in the `connect` repo (Rust)** — what datumctl invokes:

- `clap` argument parsing
- Plugin-mode detection (env-based)
- Picker UX using `inquire` (kept out of the lib so the lib doesn't force a TUI dep on consumers)
- Progress-line rendering (the `println!("  ✓ {} ({:.1}s) [{resource}]", ...)` strings) — implemented as a callback that consumes lib's `TunnelProgress` events
- `QuietTracing` (the `tracing_subscriber::reload::Handle` guard) — CLI concern, not lib
- Daemon mode (double-fork, PID files, log rotation, signal handling)
- Service-mode `run` subcommand
- Verify-endpoint output rendering

**Go plugin in the `connect` repo (separate process)** — what's *between* datumctl and the Rust binary:

- Cobra command surface
- Service-manager (`kardianos/service`) integration
- Locates and execs the Rust binary
- Reads `DATUM_*` env from datumctl; sets the right ones for the child

Lib stays Rust-only; the Go plugin is a thin process-management shell around the Rust binary, as the design already intends.

**Desktop app (`app/ui/`)** — uses `connect-lib` directly:

- Picks up the per-project `Repo` layout automatically
- Keeps `DatumCloudClient` for OAuth (unchanged)
- Keeps `HeartbeatAgent::start()` (auto-enroll) — correct behavior for the multi-project UI
- Drops any direct dependency on the cli binary (which it doesn't have today anyway)

### Suggested first step (additive against today's app/main)

1. **Extract `await_tunnel_progress` and `verify_endpoints` from `app/cli/src/main.rs` into `app/lib/`** with callback-based progress reporting. No `println!` in lib; the CLI provides closures that do the rendering. This is the only piece of today's `cli/` that's genuinely shared-library-shaped.
2. **Add `BearerSource` trait + the `ExternalTokenSource` impl** alongside the existing `DatumCloudClient`. `ProjectControlPlaneClient::new_with_bearer_source` is the entry point. This is item 1 in §Required Library Changes — additive, no migration.
3. **Land both in `app/` first**, behind the existing `--id`/picker/verify-phase flow, so the new shapes get exercised in production before the repo split. When `connect/` exists, the move is a `git filter-branch`-style extraction.

None of that requires this enhancement to be `implementable` — it's strictly additive against today's `main`.

---

@drewr — adding this so the resolutions are recorded against the enhancement itself rather than scattered across PR-130 review comments. Use the resolution table as the authoritative answer to my [previous review comment](#issuecomment-4644263879) until the enhancement body is updated.

PR #760, opened 2026-06-09
feat: add ehancement doc for pricing and offers

mattdjenkinson:
This adds a design enhancement doc for pricing and offers under billing. 

It goes with enhancement https://github.com/datum-cloud/tmp-2026-07-08-Pricing-via-Amberflo-478608ed/issues/1.


enhancements/platform/billing/pricing-and-offers/README.md
```diff
@@ -0,0 +1,438 @@
...
+  draft → publish lifecycle and the ability to switch a billing account's
+  active Offer with an audit-log entry.
+
+We will know this has succeeded when a new billing account in staging lands on
+a default Offer automatically and an invoice run for an organization with
+real usage returns a non-zero dollar subtotal.
+
+### Non-Goals
+
+- **One-time and recurring charges.** Usage-based only for v1.
```
scotwells: We should keep one-time charges and recurring charges in the v1 so the design encompasses all charge types. Otherwise we may end up with a design that works well for usage based charges and falls apart for others.

enhancements/platform/billing/pricing-and-offers/README.md
```diff
@@ -0,0 +1,438 @@
...
+            flat: "0.0000140"
+          - flat: "0.0000130" # default when no match
+```
+
+### ServicePricing fan-out
+
+service-catalog's existing `QuotaFanOut` gains a sibling `PricingFanOut` that
+emits one `ServicePricing` per priced metric, matching the `MeterDefinition`
+fan-out pattern already in place. Providers watch the single `ServicePricing`
+shape rather than parsing `ServiceConfiguration`.
```
scotwells: Is this ServicePricing CRD something that's defined by the billing service? It's not clear since the only references I see are embedded in offers.

enhancements/platform/billing/pricing-and-offers/README.md
```diff
@@ -0,0 +1,438 @@
...
+We need Amberflo to return what they owe so a downstream controller can charge
+Stripe.
+
+**How the proposal addresses it.** The `amberflo-provider` reconciles each
+`Offer` into a Product Plan (with Product Plan Items per pricing entry, honoring
+dimension filters) and each `BillingEntitlement` into a Customer-Plan
+assignment. With rates attached and a plan assigned, an Amberflo invoice run for
+the billed period returns a non-zero dollar subtotal. That subtotal is the input
+the credit ledger ([#747][747]) draws against and the future invoice-run
+controller hands to Stripe.
```
scotwells: This makes it sound like we'll be creating an invoice controller that creates invoices in stripe? Is that accurate?

enhancements/platform/billing/pricing-and-offers/README.md
```diff
@@ -0,0 +1,438 @@
...
+keeping v1 scoped strictly to usage-based billing.
+
+### Composition with quota
+
+Pricing and quota stay linked. `BillingEntitlement` activation continues to
+trigger the existing quota fan-out via the `OrganizationDefaultsReconciler` and
+`ServiceEntitlementReconciler.ensureQuotaGrants`. Pricing rides on the same
+activation event, so an Offer that costs a given amount also unlocks the right
+quota — the umbrella's requirement that quota and pricing be connected is
+satisfied without merging the two resources.
```
scotwells: I'm not understanding the relationship between pricing and quota here. Can you help me understand?

enhancements/platform/billing/pricing-and-offers/README.md
```diff
@@ -0,0 +1,438 @@
...
+controller hands to Stripe.
+
+### How Offers relate to billing models
+
+Every `Offer` carries a required `spec.billingModel`. In v1 the only valid value
+is `Usage`. Reserving the field now — rather than introducing it later — means
+`Commit`, `Subscription`, and `NonBilled` can become valid values in subsequent
+versions without a breaking schema migration. This gives an Offer an explicit,
+machine-readable answer to "how does this get billed?" from the start, while
+keeping v1 scoped strictly to usage-based billing.
```
scotwells: `Usage` is odd to me as a billing model since usage is really a type of charge. You may have a single offer that combines a one-time charge, a usage charge, and a subscription charge all into a single offer. 

What are we trying to convey with this billing model option?

scotwells (commented):
Are we able to create some real-world scenarios that capture how we might price our existing products with the proposed API surface? I'd also like to understand how service enablement relates to billing and billing accounts.

For example, what if a user has a billing account that has pricing for the DNS and Networking Services products but doesn't have any active pricing for compute in their offer?

What billing / pricing APIs did we use as competitive research when designing out these APIs? Would be good to understand how existing platforms model this information (maybe as a separate document).

enhancements/platform/billing/pricing-and-offers/README.md
mattdjenkinson: Done. Renamed spec.billingModel to spec.chargeTypes, a set that accepts Usage, OneTime, and Recurring. A single Offer can bundle all three.

enhancements/platform/billing/pricing-and-offers/README.md
mattdjenkinson: Addressed by the rename above. chargeTypes: Usage, OneTime, Recurring, reads as charge types within an Offer rather than a billing model, which is what it should have been. An Offer that combines a setup fee, a monthly platform fee, and metered consumption just declares all three in chargeTypes.

enhancements/platform/billing/pricing-and-offers/README.md
mattdjenkinson: Expanded significantly here, i'm suggesting quota gating on billing is opt in to make it easier for us to adopt? So you can add quotas for something but they don't specifically have to be tied to billing.

enhancements/platform/billing/pricing-and-offers/README.md
mattdjenkinson: `ServicePricing` will be owned by service-catalog. It's emitted automatically by two new fan-out controllers, PricingFanOut for meter-based Usage entries and ChargeFanOut for OneTime/Recurring fixed charges. The amberflo-provider reads `ServicePricing` resources to build Product Plan Items. After an Offer is published, the Offer carries a self-contained snapshot of the rates, which is why the only references you see at runtime are embedded in the Offer.

enhancements/platform/billing/pricing-and-offers/README.md
mattdjenkinson: I hadn't expanded enough on that sorry, was getting ahead. This is out of scope for this work, it'll be something we need to work out once this and crediting is in place i think.

ecv (approved):
approving on grounds sat too long & any discrepancies are best adjusted via followup iteration

PR #770, opened 2026-06-19
docs(iam): Add unified organizations enhancement design

mattdjenkinson:
## Summary

Organization `Personal | Standard` types bundle quota tier, provisioning rules, portal UX, and a loose stand-in for "individual vs business" into one immutable field, even though billing identity already lives on `BillingAccount.spec.contactInfo`. That makes orgs hard to rename, upgrade, or invite teammates into without recreating workspaces.

This adds a provisional enhancement design for collapsing to a single organization model: drop `spec.type`, add `spec.contactInfo` on the org (separate from billing contact), use opaque server-generated k8s names, migrate existing orgs to Standard-equivalent behavior, and gate full portal access on onboarding until org contact is complete.

Key outcomes documented:

- API and migration approach across milo, datum, cloud-portal, staff-portal, billing, and infra
- Implementation examples (controllers, quota policies, portal middleware, migration scripts)
- Decided defaults: deterministic `org-{hash}` names for new auto-created orgs; legacy slugs unchanged in v1; no org-to-billing contact sync

Related to https://github.com/milo-os/milo/issues/636

### DIFF
--- a/enhancements/platform/identity-and-access-management/unified-organizations/README.md
+++ b/enhancements/platform/identity-and-access-management/unified-organizations/README.md
@@ -5,11 +5,21 @@ latest-milestone: "v0.x"
 ---
 
 <!-- omit from toc -->
+
 # Unified organizations
 
 Related: [milo-os/milo#636](https://github.com/milo-os/milo/issues/636)
 
 - [Summary](#summary)
+- [What is an organization?](#what-is-an-organization)
+  - [Definition](#definition)
+  - [Why organizations exist](#why-organizations-exist)
+  - [How organizations relate to the rest of the platform](#how-organizations-relate-to-the-rest-of-the-platform)
+  - [Organization lifecycle](#organization-lifecycle-today)
+- [Product Change](#product-change)
+  - [What changes for users](#what-changes-for-users)
+  - [Why we are doing this](#why-we-are-doing-this)
+  - [Benefits](#benefits)
 - [Motivation](#motivation)
   - [Goals](#goals)
   - [Non-Goals](#non-goals)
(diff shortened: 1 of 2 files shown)


kevwilliams (dismissed):

enhancements/platform/identity-and-access-management/unified-organizations/README.md
```diff
@@ -0,0 +1,929 @@
...
+implemented, not as we intend it to end up. Where today's behavior is the thing
+this enhancement changes, that is called out explicitly and detailed in
+[Product Change](#product-change) and the [Proposal](#proposal).
+
+### Definition
+
+An organization is the top-level tenant in Datum Cloud. It is the boundary that
+owns resources, groups the people who work on them, and anchors quota and
+billing. Everything a customer does on the platform happens inside an
+organization.
```
scotwells: > It is the boundary that owns resources, groups the people who work on them, and anchors quota and billing.

Most work actually happens in projects and most resources live in projects too. Maybe it's better to say that organizations are a collection of people working on projects consuming resources from the platform?

enhancements/platform/identity-and-access-management/unified-organizations/README.md
```diff
@@ -0,0 +1,929 @@
...
+### Why organizations exist
+
+Organizations exist so the platform has a single, stable answer to four
+questions:
+
+| Question                        | Organization's role                                                                                                                        |
+| ------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------ |
+| **Who owns this resource?**     | Every project and workload belongs to exactly one organization.                                                                            |
+| **Who is allowed to touch it?** | Access is granted by membership + roles scoped to the organization.                                                                        |
+| **How much can they use?**      | Quota (e.g. project count) is granted to the organization. Today the amount depends on `spec.type` (Personal = 2 projects, Standard = 10). |
```
scotwells: This is a little misleading because quotas are enabled at the project and organization level. This makes it sound like quotas are only enabled at the organization level. 

The only quotas we set at the organization level are number of members and projects.

Maybe best to exclude this? Now that we're removing the `Type` field, we won't be setting quotas differently for each organizations without manual increase requests.

enhancements/platform/identity-and-access-management/unified-organizations/README.md
```diff
@@ -0,0 +1,929 @@
...
+drives quota tier, provisioning, validation, and portal UX, not a structurally
+different object. Collapsing these two into a single kind is the core of this
+enhancement (see [Product Change](#product-change)).
+
+An organization plays four roles:
+
+- **Workspace.** Projects and the resources inside them live here.
+- **Team boundary.** People are invited into it and assigned roles.
+- **Isolation boundary.** Resources in one org are isolated from every other org.
+- **Commercial anchor.** Quota is granted to it, and billing accounts attach to it.
```
scotwells: I'd exclude quota from this. Commercial anchor is more around billing accounts / contracts / legal documents for the organization.

enhancements/platform/identity-and-access-management/unified-organizations/README.md
```diff
@@ -0,0 +1,929 @@
...
+| **Who pays?**                   | Billing accounts attach to the org.                                                                                                        |
+
+Without this boundary, there would be nowhere to attach access control, quota,
+billing, or isolation. The organization is the join point for all of them.
+
+### How organizations relate to the rest of the platform
+
+```mermaid
+flowchart TD
+  user[User] -- membership + role --> org["Organization<br/>(spec.type: Personal | Standard)"]
```
scotwells: Aren't we removing `spec.type`? Let's focus on representing what we want to go with moving forward. I wouldn't focus too much on what exists today.

enhancements/platform/identity-and-access-management/unified-organizations/README.md
```diff
@@ -0,0 +1,929 @@
...
+does not store billing data or change the org resource model. Both values are
+the same `Organization` kind: namespace, projects, memberships. What
+`spec.type` actually controls is quota tier, how the org was provisioned,
+validation rules, portal UX, and observability labels.
+
+We want one kind of organization. Business vs individual belongs on org contact
+and billing data, not on a frozen enum. Existing Personal and Standard orgs
+should behave like Standard orgs after migration: 10-project quota (not 2),
+editable display names, team invites enabled, no Personal badge or sort-first
+treatment.
```
scotwells: This summary is pretty good, but I think we can leave out the technical details about the type field. I would focus more on what we want organizations to represent after this enhancement is completed.

scotwells (commented):
Directionally this looks right, but there's a lot of unnecessary technical detail littered throughout the document that doesn't need to be in here. 

The enhancement should focus on the product experience we want to have moving forward and less on status quo. We can briefly mention the issues around the current approach in the motivation section, but we don't need to go into too much detail on it.

enhancements/platform/identity-and-access-management/unified-organizations/README.md
```diff
@@ -0,0 +1,229 @@
...
+- [Implementation History](#implementation-history)
+
+## Summary
+
+An organization is how a customer teams up on Datum Cloud. It groups people,
+holds their projects, and ties together access and billing. Every organization
+works the same way: one kind of org, no Personal or Standard split, no
+second-class workspace created at signup.
+
+When someone signs up, they get a default organization. They finish a short
```
scotwells, line 35: > When someone signs up, they get a default organization.

Could we have the portal have them do this as part of the onboarding process? It's odd to have the backend system automatically create the organization if the UI has to be used to complete the onboarding flow anyways.

scotwells (approved):
Looks much better! Thanks Matt. One minor comment on where the organization creation happens, but doesn't hold up approval.

In terms of technical details, I don't think those belong in enhancements unless there needs to be some alignment around how it integrates with the overall system (at a very high-level). Specific work items that need to be done to complete an enhancement are better tracked through issues since they can define individual pieces that need to be completed.

enhancements/platform/identity-and-access-management/unified-organizations/README.md
mattdjenkinson: Yeah we could but I left it as automatic (but making the k8s name less personal than they currently are) so we could automatically create a billing account in the background and then all the user has to do is add a payment method instead of creating the org, creating the BA and then also adding a payment method.

scotwells: Would that save much time? The UI could just create all the resources in order after the user fills out the form. The nice part about not automatically creating it, is we can detect when users sign up and haven't finished the onboarding process much easier without leaving empty organizations / billing accounts lying around. 

We could also make the contact details of the billing account required so we know it will be provided.

mattdjenkinson: Might not save time but it's less friction. I'll try a few different options in the UI 😌

scotwells: I think you can hide the friction from the user and make it all look like one step

PR #783, opened 2026-06-26
Third-party plugin indexes & marketplaces for datumctl

scotwells:
## The problem

Today `datumctl` ships a single, curated, Datum-operated plugin catalog — and it's the *only* catalog users can discover plugins through. Third parties have no front door:

- **Community authors** can publish a binary but can't make it discoverable; the only path is a pull request into Datum's repo, which doesn't scale.
- **Enterprise platform teams** have nowhere to put internal tooling and end up emailing binaries or documenting raw install strings.
- **The raw-install escape hatch** (`owner/repo`) works mechanically but has no discovery, no curated metadata, and — most importantly — no trust signal.

## The proposal

Open the plugin ecosystem. Let **anyone** — a community author, an open-source project, or an enterprise platform team — publish their own catalog using the same manifest format Datum uses, and let users **register, search, browse, and install** across catalogs through the same friendly experience they already know.

- One command to add a catalog, with a clear one-time trust decision.
- Search and an interactive browser span every registered catalog.
- The curated Datum catalog stays the trusted, zero-configuration default.
- Clear **official** vs **third-party** trust badges everywhere a plugin appears, so provenance is never ambiguous.
- Enterprises are first-class: private internal catalogs, zero-touch provisioning, and allow-list guardrails.

## 🧭 New: an `experiences/` home for surface-level enhancements

This PR also introduces a new top-level **`experiences/`** section in the repo and files this proposal under it as `experiences/datumctl/`.

The idea worth socializing: people reach Datum through several distinct **experiences** — the Cloud Portal, `datumctl`, the API, the Kubernetes operator, CI/CD — and each is a *product in its own right*, with its own users, ergonomics, and roadmap. Organizing enhancements by the **experience they shape** (not only by the platform capability they touch) gives each of these surfaces a clear home, and makes "how does it *feel* to work with Datum here?" a first-class question rather than an afterthought bolted onto a domain feature.

- Enhancements about **an experience itself** — how it looks, feels, and behaves — live here, grouped by experience.
- Enhancements about a **platform capability** (DNS, BGP, telemetry, …) stay under their domain section, even when an experience surfaces them.

`datumctl` is the first experience to land here; Cloud Portal, API, Kubernetes Operator, and CI/CD can follow as their enhancements are proposed. Feedback on the taxonomy itself is very welcome — it's meant to scale with the platform.

## Scope

Status is **provisional**. The document focuses on the product experience and vision; implementation sequencing is intentionally out of scope.

Review the [full enhancement document](https://github.com/datum-cloud/enhancements/tree/enhancement/datumctl-plugin-marketplace/enhancements/experiences/datumctl/plugin-marketplace) as rendered markdown, and the [`experiences/` overview](https://github.com/datum-cloud/enhancements/tree/enhancement/datumctl-plugin-marketplace/enhancements/experiences) for the taxonomy.


privateip:
Did a first pass read through and overall I like the idea.  You noted in the enhancement that 

> It introduces ecosystem-management responsibility. Reserved names, trust badges, publisher verification, and impersonation defenses are ongoing concerns once third parties can publish.

In my experience the issue with reserved names is going to land fast and I think we need some way to deal with that.  I wouldn't necessarily hold up moving forward until it is completely flushed out but it will need attention quickly.

scotwells:
@privateip to me that becomes a bigger issue when we have an aggregated plugin index hub (e.g. Docker Hub, Artifact Hub, etc) where there's a central catalog that third-parties can submit plugins to. The approach laid out in this enhancement defers this to a later date and focuses on users installing indexes they trust and makes it clear that we haven't and won't validate anything in the index they're trusting. The index name is also something the user installing the index has to provide, so it's not a global trusted identity for the index. 

Let me know if there's something I'm missing with the current approach that gives pause / concern.

scotwells:
There's now a working prototype of this experience: **datum-cloud/datumctl#224**. Pull the branch to register a catalog and try `search` / `browse` / `install` end-to-end. Helpful for feeling out the manifest format and command surface while reviewing the proposal.

privateip (approved):

PR #794, opened 2026-07-02
Telemetry system design — full enhancement suite

0xmc:
This PR introduces the complete design documentation for a service provider telemetry system: a NATS + ClickHouse pipeline for collecting, storing, and exposing telemetry data (logs, metrics, traces, and flows) across the platform. It covers everything from ingest at the edge through storage, querying, export, and policy — structured as a set of cross-cutting component enhancements under enhancements/telemetry/.

#### Component enhancements:

- **ingest-pipeline** — The write path. NATS JetStream hub aggregates from Core NATS or NATS JetStream leaf nodes at each edge cluster (edge may or may not have persistent storage). The OTLP-NATS bridge receives OTLP/HTTP from the OTel Collector, extracts milo.project.id from resource attributes, and publishes one NATS message per log record or data point. ClickHouse subscribes directly to the hub via ENGINE = NATS ingest tables — no separate consumer process. Includes C4 PlantUML diagrams for production topology, edge accumulation behavior, and POC topology.
- **logs** — Log ingest schema (logs_ingest → logs_mv → logs), resource attribute contract, `datumctl logs` CLI design with streaming tail support, and tenant isolation via row policy. Covers the 7-day hot/90-day cold retention split for internal vs. tenant data.
- **metrics** — Customer-facing metrics pipeline on NATS + ClickHouse. Documents the fat ingest schema (metrics_ingest, ENGINE = NATS) with nullable type-specific columns, the MV fan-out to per-type MergeTree tables (otel_metrics_gauge, otel_metrics_sum, otel_metrics_histogram), and the NATS engine's String-not-JSON constraint and how it's handled.
- **retention** — Log TTL (hot delete at day 7 for tenants; cold move at day 7, delete at day 90 for internal) and the three-tier metrics rollup: raw → hourly (AggregatingMergeTree) → daily, with TTLs at each tier. SQL-correct rollup MVs reading metrics_ingest directly to avoid MV chaining.
- **query-layer** — Tenant-enforcing HTTP service over ClickHouse. Connects as api_reader; issues SET project_id per query; row policy is the defense-in-depth layer. Registers as a Kubernetes APIService on each project's control plane — project_id travels via X-Remote-Extra-* headers through the kube-aggregator proxy chain; the query layer does not validate tokens. Includes an explicit HTTP-hop-by-hop call chain diagram.
- **definition-policy** — LogDefinition, LogCollectionPolicy, LogRedactionPolicy, MetricDefinition, and MetricPolicy CRDs for Phase 5. Establishes the declarative contract between what a resource publishes and how it is collected, filtered, and exported. CEL-based filtering, attribute redaction, and metric aggregation rules.
- **export-policies** — ExportPolicy API design for customer-configured telemetry export to third-party platforms (Grafana Cloud, Datadog, etc.) over OTLP or Prometheus Remote Write. Drops the metricsql field entirely. Proposes resourceSelectors (Kubernetes-style namespace/label/kind selectors) as the backend-agnostic selection mechanism.
- **network-telemetry** — Three collection paths for Phase 3: gNMIc (streaming OpenConfig telemetry → telemetry.network.<device_id>), SNMP via the OTel Collector snmpreceiver (→ telemetry.metrics.internal), and Akvorado (NetFlow/IPFIX/sFlow → dedicated ClickHouse instance). Documents why Akvorado's internal Kafka is an accepted exception to the "minimize database engines" principle.
- **operations** — Alert definitions and runbooks for the ingest pipeline (TelemetryBridgeDropRateHigh) and ClickHouse storage (ClickHouseHotTierLow), plus a staging validation checklist for properties the design depends on but that the platform doesn't enforce automatically (e.g. confirming the row policy throws rather than silently returning empty results when project_id is unset).

#### Notable design decisions

- The telemetry stack is framed as generic service provider infrastructure (milo-os/telemetry), deployable by any service provider, not Datum-specific. OTel resource attribute keys use the milo.* namespace.
- Billing-relevant telemetry is shed last under capacity pressure; operational telemetry (gNMIc, system logs) is shed first.
- project_id is the only guaranteed tenancy key flowing through the system. All hierarchy resolution stays in the control plane.
- The edge design support both NATS Core and NATS JetStream at the edge depending whether or not the edge has storage.


enhancements/telemetry/network-telemetry/README.md
scotwells, line 1: Can we move this under `enhancements/networking/telemetry`? This should really be about how our networking layer plugs into the telemetry system. Everything under `enhancements/telemetry` should be about the core telemetry platform, not how services integrate with it. 

I also think we should have this focus on what telemetry consumers and service providers can expect from the networking service. We can move implementation details on how we retrieve that telemetry and feed it into the telemetry system as a design document in the networking repo responsible for scaffolding the telemetry collection.

scotwells (commented):
Thoughts on keeping everything in this repo as the product facing features that consumers and service providers would expect from the telemetry platform? 

It helps to keep design documentation on how we implement the system separate from how end-users expect it to work. The internal implementation can change without impacting end-users, so no need to update the product focused enhancements when those details change. 

I'd recommend we keep design documentation in the milo-os/telemetry repo for core telemetry infrastructure, datum-cloud/compute for generic compute observability, datum-cloud/unikraft-provider for unikraft specific observability, etc.

PR #807, opened 2026-07-08
docs: add enhancement for reducing PolicyBinding tuple fan-out

mattdjenkinson:
## Summary

This adds a provisional IAM enhancement proposing that a PolicyBinding write a single role-assignment tuple instead of one OpenFGA tuple per (subject, permission).

The organization owner role grants every permission across the whole resource hierarchy, so a single owner binding expands to 351 tuples. That expansion sits on the signup critical path: a new user cannot use their organization until the owner grant has propagated, and the grant is 351 tuple writes behind a reconciler that processes bindings one at a time.

The proposal moves the fan-out out of tuple data and into the authorization model. A binding writes one "subject has role R on object" tuple, and the model expresses which permissions role R grants. The model is computed once and only changes when roles or protected resources change, so the per-binding cost drops from N tuples to one.

This PR is the design only. Implementation lives in [datum-cloud/auth-provider-openfga](https://github.com/datum-cloud/auth-provider-openfga) and would follow once the approach is agreed.

## Why we need it

Signup latency on staging is dominated by IAM propagation, and the owner grant is the single largest contributor. From controller and OpenFGA logs across three real signups on the same day:

| Signup      | Org owner ready | Project owner ready |
| ----------- | --------------- | ------------------- |
| `org-kzjnm` | +59s            | +82s                |
| `org-tbsxl` | +25s            | +47s                |
| `org-drc28` | +12s            | +37s                |

The 12s to 59s spread for the same operation tracks OpenFGA write latency and reconciler queue pressure, not user behavior. Supporting numbers from the same window: OpenFGA `Write` p99 about 990ms, `Check` p99 about 10ms, and `policybinding` reconcile p99 up to 3.4s on a bad day. The tuple count was confirmed directly from the teardown log line (`Successfully deleted tuples for PolicyBinding ... "tupleCount": 351`).

## Test plan

- [ ] IAM approvers review the proposed model change and migration path
- [ ] Confirm Check-equivalence approach is sufficient to catch authorization regressions before implementation starts
- [ ] Resolve the two open questions (role scoping, group composition)

## Notes for reviewers

The document is intentionally `provisional`. The two open questions are marked with `UNRESOLVED` blocks. Please push back on the migration ordering in particular, since the safety of the rollout depends on both representations resolving to identical Check results while they coexist.

### DIFF
--- a/enhancements/platform/identity-and-access-management/policybinding-tuple-fanout/README.md
+++ b/enhancements/platform/identity-and-access-management/policybinding-tuple-fanout/README.md
@@ -1,0 +1,241 @@
+---
+status: provisional
+stage: alpha
+latest-milestone: "v0.x"
+---
+
+<!-- omit from toc -->
+
+# Reduce PolicyBinding tuple fan-out
+
+Related: [datum-cloud/auth-provider-openfga](https://github.com/datum-cloud/auth-provider-openfga)
+
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Where the time goes](#where-the-time-goes)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [How it works today](#how-it-works-today)
+  - [The change](#the-change)
+- [Design Details](#design-details)
+  - [Authorization model](#authorization-model)
+  - [Policy reconciler](#policy-reconciler)
+  - [Migration](#migration)
+- [Risks and Mitigations](#risks-and-mitigations)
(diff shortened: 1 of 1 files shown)


scotwells:
We can keep this enhancement in the OpenFGA provider repository since it's an implementation specific enhancement that's isolated to that repo. This repo should only be used for product level enhancements and higher level cross service concerns.

We'll have to performance test the impact this would have on the authz system. When we improved performance earlier this year, we moved to writing more tuples because it gives you better read time (authz check) performance at scale.

PR #820, opened 2026-07-14
Service-contributed UI for the cloud portal and service-provider portal

scotwells:
## Why

Every new service we add — compute, networking, DNS, telemetry, and whatever comes next — needs a presence in two separate portals: the customer-facing cloud portal, and the internal portal our own team uses to operate and support customers. Today that means the service team's work has to land as hand-built pages inside someone else's shared codebase, twice, on someone else's release schedule.

That gets worse as we add more services, not better. Every service's change is small and specific to that one service, but they all compete for the same shared route files, navigation arrays, and resource modules — so growth means more merge conflicts, more coordination, and more release trains coupling unrelated services together, purely because of where the code has to live.

## What this proposes

A service team declares, once, in the same catalog configuration they already use for billing and quota, that their service ships a UI — for the cloud portal, the service-provider portal, or both. The platform turns that declaration into a runtime plugin: the portal discovers it, loads it, and gates it behind the same entitlement and permission checks a built-in page would use. No PR against a portal repo, no portal deploy, no waiting on anyone else's release.

Concretely, this means:

- **A service ships its own UI, on its own schedule.** Adding or updating a service's portal experience is a normal deploy from the service's own repository, not a change to the portal.
- **The two portals get their own version of this**, because they serve very different audiences: the cloud portal shows a customer their own enabled services; the service-provider portal lets our own ops staff see a service's contribution while investigating a customer's account — which needs its own, stricter set of checks (see the doc for the security model).
- **It borrows a proven pattern.** The design follows [OpenShift Console's dynamic plugins](https://github.com/openshift/console/tree/master/dynamic-demo-plugin) — this isn't a novel, unproven mechanism.
- **A working prototype already exists** for the cloud-portal side — see [datum-cloud/cloud-portal#1353](https://github.com/datum-cloud/cloud-portal/pull/1353).

## What this doesn't do (yet)

This is scoped to first-party, platform-vetted service teams — not a public plugin marketplace. The service-provider portal side is designed in this doc but not yet built; it ships as a defined follow-on once the cloud-portal path is stable.

## Related work

[datum-cloud/staff-portal#562](https://github.com/datum-cloud/staff-portal/issues/562) proposes extracting a generic Milo Operator Console from staff-portal — the kind of generalized host this design's service-provider portal side would need in order to load plugins at all. Worth reading together: that issue shapes what the console *is*, this proposal shapes how services *plug into* it.

## Full proposal

Read the [full design proposal](architecture/experiences/web-portals/plugin-system/README.md) for the resource shapes, the manifest contract, the security model for both portals, local development, and open questions still worth discussing.

This is a draft to gather feedback on the *idea* — whether this is the right shape for how services should contribute UI to our portals — before locking in the finer implementation details.

### DIFF
--- a/architecture/experiences/web-portals/plugin-system/README.md
+++ b/architecture/experiences/web-portals/plugin-system/README.md
@@ -47,6 +47,9 @@ This shows up identically in both portals today: the consumer portal carries a h
 
 ## Architecture overview
 
+> [!IMPORTANT]
+> Everything from here down — resource kinds, API groups, field names, the manifest shape, extension-point types — is illustrative of the design's *shape*, not a final contract. Expect the exact schema to change as the design is reviewed and implementation starts.
+
 ```mermaid
 sequenceDiagram
     participant ST as Service team
--- a/architecture/experiences/web-portals/plugin-system/README.md
+++ b/architecture/experiences/web-portals/plugin-system/README.md
@@ -1,0 +1,349 @@
+---
+status: provisional
+stage: alpha
+latest-milestone: "v0.1"
+---
+
+# Service-Contributed UI for the Cloud and Service-Provider Portals
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


mattdjenkinson (approved):
I _like_ this and think you've covered my concerns. I think i need to see it working to really get my head around it. 

Are we going to be putting too much load on the system with every client constantly watching for updates? I think we'll need to be smart about where in the app we do the initial fetching so we're not adding loads of blocking time or at least make sure we have some nice loading state.

scotwells:
@mattdjenkinson the watch here would be server side, between the cloud portal and the Kubernetes control plane, not client side. The client side would just request assets, the server side is what resolves assets that are available based on plugins that are registered.

kevwilliams (approved):

Issue #800, opened 2026-07-07
Project suspension & lifecycle controls


scotwells:
A safe, reversible way to stop a bad actor without destroying an innocent customer's work — and to restore them intact if they were wrongly flagged.

Today the only way to stop an abusive project is to delete its workloads: destructive, irreversible, and incomplete. Datum can already suspend a *user* and a *billing account*, but has no reversible way to stop a **project**. Project suspension fills that gap — the core principle is **"suspension is never deletion."**

## Product goals

- **Stop abuse now, reversibly.** Trust & safety (or an automated policy) can freeze a project in one action — new work is blocked and running work is paused, not deleted.
- **Make wrongly-flagged customers whole.** Reinstatement returns a project to exactly where it left off, with nothing lost. Abuse holds are operator-reviewed; billing holds are customer self-remediable.
- **Behave consistently everywhere.** A single integration contract lets every managed service honor suspension the same way, so the experience doesn't vary service to service.
- **Keep customers informed.** A suspended project shows the reason and a clear path to recover; the suspend → pause → reinstate → resume sequence is visible on the project's timeline.

## Scope

**Compute is the launch integrator** — pausing a project's instances losslessly is the launch-blocking need. The contract is designed so other services (networking, DNS, and beyond) adopt it later. Detailed per-service and neighboring-system work (billing-account cascade, provider-initiated service-scoped suspension) is deferred to separate enhancements.

## Proposal

Full design: **[Project Suspension & Lifecycle Controls](https://github.com/datum-cloud/enhancements/blob/enhancement/project-suspension/enhancements/platform/project-suspension/README.md)** (PR #821).

Delivery is tracked across the datum-cloud and milo-os organizations in the sub-issues below.

JoseSzycho:
Project Suspensions are working correctly in staging.

Staff Users UI as backend service is working correctly.

Right now promoting the feature to production, so we can test it agains the edge cluster, in order to tests the compute pods gets correctly delete/reinstated (staging does not include an edge cluster for testing this).

Cloud portal UI is soon to receive the update, the work is already in process.

JoseSzycho:
The ProjectSuspension feature is now deployed to production, and working as expected.

Continuing with:
- https://github.com/milo-os/milo/issues/721

JoseSzycho:
all sub issues has been tackled. They new functionality is fully tested in prod/staging, and has been working correctly for weeks.

PR #821, opened 2026-07-14
Project suspension & lifecycle controls (product direction)

scotwells:
> [!IMPORTANT]
> Review the full proposal as a rendered document: [Project Suspension & Lifecycle Controls](https://github.com/datum-cloud/enhancements/blob/enhancement/project-suspension/enhancements/platform/project-suspension/README.md). The diff view splits prose and diagrams across files and does not render the Mermaid diagrams; the rendered README is the intended reading experience.

## What this is

Introduces the **Project Suspension & Lifecycle Controls** enhancement: a reversible, non-destructive way to *pause* an entire project — stop everything it is running and block new work — without deleting any data, and resume it intact when the project is reinstated.

Today, the only lever a trust-and-safety operator has to stop an abusive project is to delete its workloads: destructive, irreversible, and incomplete. Milo already has reversible-stop primitives for users and billing accounts, but none for a project. This proposal fills that gap with a first-class `Suspended` state on the project, declared once in the control plane and honored everywhere. The core principle is **"suspension is never deletion."**

## Scope of this PR

The goal here is **alignment on product direction and high-level architecture** — not a final, implementation-ready spec. This document establishes:

- The product behavior: what suspension does, the reversible lifecycle, and the consumer/operator experience.
- The architectural shape: project suspension as a **platform primitive** that many triggers (operator, fraud, billing, compliance) depend on, and that every managed service honors through a single integration contract.
- Clear boundaries: what belongs to this primitive vs. what each service and neighboring system owns.

**This proposal will change as we learn more about the integration.** Deliberately out of scope here, to be fleshed out in **separate design documents in the individual service repositories**:

- Per-service pause/resume mechanics — e.g. the compute instance snapshot/suspend/resume primitive ([tracked in datum-cloud/compute#182](https://github.com/datum-cloud/compute/issues/182)).
- The billing service's integration (delinquency detection and the cascade across billing-account bindings).
- Provider-initiated, service-scoped suspension (a service-catalog concern).
- Activity-service rendering of the suspend/reinstate timeline.

The service-specific examples in this document are **representative of what the platform enables**, not a specification of how any service implements it.

Tracking issue: [Project suspension & lifecycle controls (#800)](https://github.com/datum-cloud/enhancements/issues/800)


ecv (dismissed):
nice

jacobsmith928 (dismissed):
This looks good to me @scotwells my only suggestion is that we consider the fraud and abuse possibly separate things. It may not matter, but they are related but different signals and I wonder if calling that out in the top diagram would be useful long term?

scotwells:
@jacobsmith928 good call out. I agree fraud and abuse should be separate triggers. I'll push some updates shortly.

scotwells:
@jacobsmith928 updated!

drewr (approved):
Everything makes sense. No major feedback right now.

PR #831, opened 2026-07-17
docs: tighten project-ready ceiling to 10s

ecv:
## Summary

Follow-up to datum-cloud/enhancements#829. Tightens the project-ready ceiling from 15s to 10s and makes one rule explicit: outliers and tail defects do not hold a ceiling open.

The 15s number was held back from 10s because the observed p99 tail sits above it. That got the derivation backwards — the ceiling is what a customer could reasonably expect (project creation is an interactive flow and should feel near-instant), and the tail is a defect tracked against the ceiling, not a constraint on it. Healthy p95 is single-digit seconds, so 10s is achievable.

Related to datum-cloud/infra#3692.

## Test plan

- [x] Table row and Design Details prose updated; no other content touched

### DIFF
--- a/enhancements/platform/convergence-slos/README.md
+++ b/enhancements/platform/convergence-slos/README.md
@@ -58,7 +58,7 @@ The initial set of user-facing convergence paths and their thresholds:
 
 | Convergence path | Signal that it converged | Reasonable-expectation threshold (proposed) |
 |---|---|---|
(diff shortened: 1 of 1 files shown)


ecv:
_provocative_ PR

scotwells (approved):

PR #852, opened 2026-08-07
feat: add datumctl vpc plugin enhancement

privateip:
Introduce a kubectl-modeled plugin for VPC lifecycle management with
create, get, describe, and delete actions. Initial scope is IPv6-only
(ULA); GUA and IPv4 support are deferred.

Includes enhancement design doc and datumctl index entry.


privateip:
Implements https://github.com/datum-cloud/galactic/issues/200

ecv (approved):

Issue #792, opened 2026-07-01
DNS64 Support for Multi-Tenant DNS Service


privateip:
### High-Level Summary

Add DNS64 support to Galactic VPC's multi-tenant DNS service. When a tenant's IPv6-only compute instance queries a hostname that has no AAAA record, the resolver synthesizes one from the destination's A record using the tenant's NAT64 prefix, so the instance gets a routable IPv6 address that the NAT64 gateway can translate back to IPv4 on the way out. Without this, IPv6-only instances simply fail to resolve IPv4-only destinations.

### Motivation

IGW is adding NAT64 support for IPv6-only compute instances reaching IPv4-only destinations, but NAT64 alone is useless if the instance never gets an address to send traffic to in the first place. An IPv6-only workload resolving an IPv4-only hostname today gets no AAAA record and no path forward. The application either hard-fails the connection or falls back to a resolver outside Datum's control, silently bypassing the NAT64 gateway and any per-tenant policy attached to it. DNS64 closes that gap. It's the standard pairing (RFC 6147) for any NAT64 deployment, and every IPv6-only tenant onboarding onto Galactic VPC will hit this immediately. This is not an edge case; it is table stakes for IPv6-only compute to function on the public internet at all.

### Goals

- The multi-tenant DNS service synthesizes AAAA records from A records for queries that return no native AAAA, using the tenant's assigned NAT64 prefix.
- Synthesis is scoped per-tenant private zone, consistent with existing per-tenant DNS isolation. One tenant's synthesis behavior never leaks into another's.
- The DNS service reads the NAT64 prefix from IGW as the source of truth; no NAT64 prefix is independently configured or duplicated in the DNS service.
- Tenants can disable DNS64 synthesis per zone for workloads that need to observe the real absence of AAAA (e.g. IPv6-only-strict validation, dual-stack detection tooling).
- Synthesized responses are marked/logged distinctly from native AAAA responses so tenant-facing debugging tools can show a client whether an address was synthesized.
- DNS64 behavior degrades safely if the IGW/NAT64 prefix is unavailable: queries return no AAAA (matching pre-DNS64 behavior) rather than serving stale or incorrect synthesized addresses.

### Non-Goals

- No DNS64 responder or synthesis logic is added at the IGW. IGW remains the source of the NAT64 prefix only.
- No support for excluding specific destination prefixes from synthesis (DNS64 "exclusion lists" per RFC 6147 §5.1.4) in this phase. Global per-zone on/off only.
[...]

Issue #791, opened 2026-07-01
Internet NAT64 gateway for Galactic VPC


privateip:
### High-Level Summary

Add NAT64 to Galactic VPC's Internet Gateway (IGW): a stateful IPv6-to-IPv4 translation service that lets IPv6-only compute instances reach the legacy IPv4 internet without a dual-stack NIC, a bump-in-the-wire NAT appliance, or any change to the instance's networking config. Tenants provision IPv6-only instances as they do today; outbound traffic to IPv4-only destinations is translated at the IGW and returned untranslated to the originating instance.

### Motivation

Galactic VPC's IPv6-only instance type has no path to the IPv4 internet today, and a meaningful chunk of the internet — legacy APIs, vendor endpoints, CDNs without AAAA records — is still IPv4-only. Tenants adopting IPv6-only compute to cut address management overhead currently have to choose between reverting to dual-stack (defeating the point) or running their own NAT64 translator inside the VPC, which pushes an operational burden onto the tenant that we're better positioned to run centrally and correctly. This is table-stakes connectivity for any provider offering IPv6-only compute; without it, IPv6-only is a curiosity, not a real deployment option. RFC 6146 defines the mechanism — nothing novel required on the standards side, this is about integrating it cleanly into our SRv6/IGW data plane and making it operate at Datum scale.

### Goals

**Phase 1 (ships first — closes the gap; nothing in the fabric provides IPv4 egress today):**

- IPv6-only instances can reach IPv4-only destinations on the internet without any tenant-side configuration, NAT64 prefix handling, or client-side awareness that translation is happening
- NAT64 translation reuses the exact operational pattern our existing NAT66 tier already proves out in production: node-local, sharded translation gateways, with a tenant VRF's traffic routed to an assigned shard at attachment time
- Per-tenant session isolation is enforced via the SRv6 Argument field on the shard's NAT64 locator, so no tenant's translation table or session load can affect another tenant's
- Tenants get visibility into NAT64 session counts and can see when they're approaching a documented per-tenant session limit, with the limit enforced before it degrades other tenants
- This ships and is validated alongside DNS64 ([#792](https://github.com/datum-cloud/enhancements/issues/792)): an IPv6-only instance resolving an IPv4-only name via DNS64 must successfully connect through this NAT64 path end to end

**Phase 2 (deferred until Phase 1's one real limitation — a shard restart breaks the sessions it was servicing — is a demonstrated operational problem, not built speculatively ahead of that evidence):**

[...]

PR #879, opened 2026-09-11
docs: Add NAT64 gateway design for VPC networks

### LINKED ISSUE #866: Research & Design: NAT64 for IPv6-Only Instances on Galactic VPC
### Research & Design
This issue tracks research and discovery for #791 (NAT64 for IPv6-Only Instances on Galactic VPC).

### Research Goals

- Determine how RFC 6146-compliant stateful NAT64 translation integrates cleanly into Galactic's SRv6/IGW data plane (the edge XDP NAT+LB gateway datapath — see `internal/plumbing/ebpf/edge{prog,map,attach}` and [ARCHITECTURE-GATEWAY.md](https://github.com/datum-cloud/galactic/blob/main/docs/agents/ARCHITECTURE-GATEWAY.md)), versus a separate translation component alongside it.
- Evaluate how to advertise NAT64 translation via an anycast SID so any IGW node in-region can service a session, and how that fits the existing Active-Active BGP local-preference placement model.
- Design the regional session-state mechanism so a session survives failover between IGW nodes within a region without breaking the connection — including what state needs to be shared/replicated and where it lives.
- Design per-tenant session isolation using the SRv6 Argument field on the shared NAT64 locator, so no tenant's translation table or session load can affect another tenant's.
- Define per-tenant session limits: default value, enforcement point (before it degrades other tenants), and how tenants get visibility into current session counts and how close they are to the limit.
- Identify what's needed for end-to-end validation together with DNS64 (#792) — an IPv6-only instance resolving an IPv4-only name via DNS64 must successfully connect through the NAT64 path end to end.
- Confirm scope boundaries hold up technically: outbound-only (no inbound NAT64), no cross-region session failover, single Datum-managed translation prefix (no per-tenant customization), no changes to DNS64 resolver behavior.

### Deliverables

- Design doc covering: SRv6/IGW data-plane integration point, anycast SID advertisement, regional session-state architecture and failover behavior, per-tenant isolation via the SRv6 Argument field, and session-limit enforcement/visibility model.
- Recommendation on where translation logic lives (existing edge XDP gateway program vs. new component) with tradeoffs.
- Proposed API/CRD surface (if any) for tenant-facing session visibility and limits.
- Test/validation plan for end-to-end DNS64 + NAT64 connectivity.
- Open questions and risks flagged for the implementation enhancement.

### Additional Context

- Parent enhancement: #791
- Related: #792 (DNS64)
- Relevant docs: [ARCHITECTURE-GATEWAY.md](https://github.com/datum-cloud/galactic/blob/main/docs/agents/ARCHITECTURE-GATEWAY.md), [ARCHITECTURE-ROUTER.md](https://github.com/datum-cloud/galactic/blob/main/docs/agents/ARCHITECTURE-ROUTER.md)
- RFC 6146 (Stateful NAT64)

privateip:
## Summary

Galactic VPC's IPv6-only instances have no path to the IPv4 internet today.

This proposes NAT64 translation delivered by generalizing the existing NAT66 binary and CRD into one combined implementation, rather than building a second, near-duplicate one.

Anycast advertisement and a regional session store are explicitly rejected for now, in favor of shipping the proven node-local pattern first.

Scope stops at translation and local metric collection. Emitting that data anywhere is out of scope for this document.

## Test plan

- [ ] Markdown renders correctly on GitHub, with all internal links and the table of contents resolving

Resolves #866
Related to #791, #792

🤖 Generated with [Claude Code](https://claude.com/claude-code)


privateip:
Reconciled against the implementation ([galactic#525](https://github.com/datum-cloud/galactic/pull/525), [network#22](https://github.com/datum-cloud/network/pull/22)). Corrections are marked in place with what the original got wrong, so a reader can tell a design decision from a discovery.

**Wrong as written:**

- Design Details claimed NAT66 was already in production, twice, while Risks correctly said the opposite. It isn't. That removes the migration cost from the CRD rename, but it also cuts against the case for combining — which leaned on consolidating onto a proven binary — so the argument is restated on grounds that hold.
- "RFC 6146-compliant" was larger than intended. ICMP translation and fragment handling are now explicit non-goals; scope is TCP and UDP, and PMTUD across the translator does not work.
- The function code does not select the family. One SID serves both, and dispatch reads the inner destination. Enabling NAT64 adds no SID and no per-tenant action.

**Missing:**

- Verifier complexity — the "distinct translation functions" mitigation made it worse, since both paths force-inline into one program. Separate programs behind a tail-call dispatcher instead.
- The LRU session table cannot keep its own count accurate, so a per-tenant ceiling degrades into a dead shard without a userspace resync.
- Nothing in this design makes a shard's public IPv4 address reachable. Most likely way a first deployment fails.
- The CRD rename is cross-repo and has to land first.

**Validation** is restated against outcomes. Steps 3–5 are met at the datapath level, with both translated checksums verified against independent recomputes. Steps 1–2 cannot be met in containerlab — its mesh is IPv6-only, so NAT64 there would translate correctly and then blackhole, indistinguishable from broken. Two checks added that the original lacked.

Also records the pre-existing NAT66 forward-path defect the new regression coverage found: the decapsulated packet never got a link header written, so its EtherType was stale packet bytes and it reached no protocol handler. Relevant here because the half described as proven had never had its forward path exercised end to end.

